PR #68, opened 2018-12-10
Make server evaluation handling more generic

mattflax:
This PR separates the evaluation handling from the RREController class, allowing us to use Spring profiles to change how the evaluation request is handled, and how the evaluation is built from the data in the request. The default handler will still read the entire evaluation out of the incoming HTTP request.

I have included an evaluation handler which will read the evaluation from a URL passed in the evaluation request, allowing the evaluation file to be uploaded to a central location or file server. This allows larger evaluation files to be handled (although the 5-second refresh rate on the front end may be problematic). It can be run using 

    java -Dspring.profiles.active=url -jar rre-server.jar

### DIFF
--- a/rre-core/src/test/java/io/sease/rre/persistence/impl/JsonPersistenceHandlerTest.java
+++ b/rre-core/src/test/java/io/sease/rre/persistence/impl/JsonPersistenceHandlerTest.java
@@ -9,6 +9,7 @@ import org.junit.Test;
 import org.junit.rules.TemporaryFolder;
 
 import java.io.File;
+import java.io.FileWriter;
 import java.util.HashMap;
 import java.util.Map;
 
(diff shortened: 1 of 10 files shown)


agazzarini:
The PT looks good to me @mattflax. 
Please confirm if I got the thing: when the evaluation data is generated, the report plugin 

- uploads the evaluation data in a remote server
- sends the URI where to fetch the data to the RREController

I'm completely ok with this, just wondering about the functional scenario: what is the advantage of such decoupling? Evaluation data persistence managed outside the RREController (where it is transient at the moment)?

mattflax:
Hi Andrea,

Thanks for taking a look at this. You are correct - the report sends a URI for the file to the RREController, which then reads the evaluation data from the file. It doesn't have to be a remote location if the plugin and the controller are on the same box, but a URI seemed more flexible than a file path.

I did this as a first step towards connecting the server to Elasticsearch, and reading the evaluation data from there. Of course, there is potentially a lot more data in ES, so I may have to work out some ways of filtering it, possibly adding more controls to the dashboard.

I'm not sure how best to structure the ES implementation - whether it should be a plugin to RREServer, or if I should put it with the main code. If you have any thoughts on that, please let me know.

mattflax:
The query set we're working with at this stage has ~800 queries in it, and the ES persistence handler stores one document per version. If we have, for example, 4 configuration sets, that means pulling 3200 documents from ES and using them to build an evaluation that can be used by the front end. That evaluation JSON can get very large, especially if the user has chosen to include many fields in the stored query results, and the dashboard display can take time to update its model as well. So I'm thinking of filtering by topic, query group, version, perhaps limiting the display metrics as well, to reduce the amount of data being passed about.

I had originally thought of using Kibana to display the metric averages, but it's not good if you need to drill down to the query level. You need to define one visualisation per entry (overall, topic, query group, etc.), which is okay when you only have a few items to visualise, but adding 800 visualisations to allow each query to be tracked is not practical.

agazzarini:
Great, sounds good. Consider that we built the RRE Server with a poor knowledge of front-end technologies (and I'm sure you already got it, looking at the code). 
So if you think something could be improved by changing or even rewriting that module absolutely feel free to propose.

PR #73, opened 2019-01-30
Allow parameters to be passed to generic metrics

mattflax:
This PR adds the option to pass configuration parameters to generic metrics - for example, PrecisionAtK. This means that we don't need one class definition for every concrete implementation of the generic metrics.

It requires that the generic metric have a constructor with `@JsonProperty` annotations, matching up with the properties passed in the config - I've made this change in the PrecisionAtK class in the patch.

The configuration is passed from Maven via the pom.xml. In addition to the `metrics` block, there is a new, optional, `parameterizedMetrics` block, like so:

```
<metrics> ... metrics here </metrics>
<parameterizedMetrics>
  <metricName>  <!-- Can be anything, but must be unique -->
    <!-- Name of the metric class to instantiate -->
    <class>io.sease.rre.core.domain.metrics.impl.PrecisionAtK</class>
    <!-- Constructor parameters, matching @JsonProperty annotations -->
    <k>5</k>
  </metricName>
</parameterizedMetrics>
```

I hope this sounds sensible, and look forward to getting your feedback!

### DIFF
--- a/rre-core/src/main/java/io/sease/rre/core/Engine.java
+++ b/rre-core/src/main/java/io/sease/rre/core/Engine.java
@@ -7,8 +7,8 @@ import io.sease.rre.Field;
 import io.sease.rre.Func;
 import io.sease.rre.core.domain.*;
 import io.sease.rre.core.domain.metrics.Metric;
+import io.sease.rre.core.domain.metrics.MetricClassManager;
 import io.sease.rre.persistence.PersistenceConfiguration;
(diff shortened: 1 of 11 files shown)


agazzarini:
Hi @mattflax my feedback is absolutely positive! Let me have a quick review and I'll come back to you. 

Thanks
Andrea

agazzarini:
Many thanks @mattflax, great addition!

PR #76, opened 2019-02-13
FMeasure and Recall at K

worleydl:
This PR adds FMeasureAtK and RecallAtK which can be used in the new parameterized metrics that were recently added.  I included some test cases as well as a new test case for PrecisionAtK.


agazzarini:
Many thanks @worleydl !

Issue #78, opened 2019-02-25
Solr does not allow unquoted JSON DSL in query template


mattoscln:
When including JSON DSL in a Solr query template, it must be quoted or is not added to the query correctly.

**To Reproduce**
1. Add a JSON parameter to a Solr query template:
```
{
  "json": {
    "query": {
      "lucene": {
        "query": "$query"
      }
    }
  }
}
```
2. Run the evaluation process.
3. Observe that the query details have not been added to the SolrQuery (eg. the query request looks like `/select?fl=field1 field2&json=&rows=20`

**Expected behavior**
Included JSON DSL should not require quotes around the content - this will make it harder to write and increase the number of potential bugs in the query where JSON is in use.

agazzarini:
Closed with https://github.com/SeaseLtd/rated-ranking-evaluator/pull/79

PR #79, opened 2019-02-25
#78: Modifications to handle unquoted JSON string in query template.

mattflax:
Modifications to allow JSON to be used without quotes in the Solr query template.


mattflax:
Sorry, misread the issue number. This refers to issue #78 - hopefully will show up correctly!

agazzarini:
Thanks @mattflax !

Issue #59, opened 2018-10-06
Parallel query execution 


agazzarini:
When the ratings file is huge, and a lot of queries need to be executed as part of the evaluation process, it would be great to have some kind of parallelism.

PR #80, opened 2019-03-19
#59: Parallelize query evaluations

mattflax:
This PR adds the option to parallelize the query evaluation, as per issue #59. There are three choices:

- run query evaluation synchronously, where evaluation blocks the main thread (current behaviour);
- run query evaluation asynchronously, where evaluation runs in parallel to the main thread but each versioned query is run on the same thread;
- run query evaluation asynchronously, with each versioned query on its own thread (we've found this to be slower than the basic async evaluation, though it may be viable on a more powerful server).

The default is to use the second option, with 4 threads. Configuration is passed via the pom.xml - the archetypes have been modified to include the default configuration values.

(I also added a caching template manager, which seemed easier to pass around than reading the template every time.)

I hope this approach makes sense - obviously I'd appreciate any feedback!


agazzarini:
Impressive work @mattflax, let me have a better look in the weekend

PR #81, opened 2019-04-03
Expected Reciprocal Rank (ERR) metric

binarymax:
This adds the ERR metric capability.  The metric is described in the paper "Expected Reciprocal Rank for Graded Relevance":  http://www.olivier.chapelle.cc/pub/err.pdf 

As implemented in this pull request, it uses the default gain function of ((2^value)-1)/(2^max).  No topic diversity capability is currently implemented.  The default cascade model of (1/rank) is also used.  This metric will be extended in the future to allow variations on the gain function, cascade model, and a multiple diversity setting.

### DIFF
--- a/rre-core/src/main/java/io/sease/rre/core/domain/metrics/impl/ExpectedReciprocalRank.java
+++ b/rre-core/src/main/java/io/sease/rre/core/domain/metrics/impl/ExpectedReciprocalRank.java
@@ -1,0 +1,92 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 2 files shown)


agazzarini:
Great @binarymax! Thanks a lot!

agazzarini:
@binarymax it would be great if you could add this metric with short description here https://github.com/SeaseLtd/rated-ranking-evaluator/wiki/Evaluation%20Measures#available-measures

binarymax:
Hi @agazzarini , happy to help with documentation but I don't have access to contribute to the wiki.  I added a description to the README though.  If you grant me wiki editing access I'll happily add much more documentation!

agazzarini:
Hi @binarymax let me check how it works here in Github, I'm not so good at it :D. In the meantime thanks a lot, I'll merge the other PR

binarymax:
Hi @agazzarini looks like you can open access to the wiki like this: https://help.github.com/en/articles/changing-access-permissions-for-wikis 
...unfortunately it lets anyone change the wiki and not just specific people.  Maybe this will be OK for you?  I know at least me, @mattflax and @worleydl would like to contribute to the documentation.

agazzarini:
Hi @binarymax the option is slighty different but should have the same effect. Please give a try and let me know if you can edit now

binarymax:
@agazzarini yes I can edit the wiki now!  Thank you!

agazzarini:
Great to hear! You're warmly welcome. 

BTW I'm so sorry I won't be able to attend the Haystack conf, so we have to pospone a beer to the next Haystack EU 🍺

binarymax:
We will give Alessandro an extra beer and he can bring it to you :)

PR #98, opened 2019-07-30
ES Persistence plugin - connectors for ES6 & ES7

mattflax:
I've modified the persistence plugin to allow connections to both ES 6.x & 7.x. 

It uses the Elasticsearch 7.2.0 client, which is not especially backwards compatible - it is forced to used the lower-level calls to deal with ES6, since the connector adds a bunch of request parameters which cause `Bad Request` errors on many ES6 iterations.

It does not work with ES5.


agazzarini:
Hi @mattflax sorry I completely forgot this PR...and it seems quite old. Before starting a review, can I ask you if you think it is still valid?

mattflax:
No worries. I'd forgotten about it as well!

I don't think the current ES persistence plugin will work with ES7, and this adds that functionality, so I do think it is still useful. It's good to have backwards compatibility with ES6 too, and this contains that as well, so I think it is still valid.

(Sorry to add to your workload!)

PR #110, opened 2020-03-12
Add support for basic authentication to external ES platform

telendt:
Currently external Elasticsearch setup does not support basic auth (encoding user/password into `hostUrl` does not work - tried that). Elasticsearch REST client documentation shows how this should be done:
https://www.elastic.co/guide/en/elasticsearch/client/java-rest/5.6/_basic_authentication.html

So I've extended `index-settings.json` with optional `user` and `password` params that are used for setting up basic auth (if present).


telendt:
Also, I think it would be helpful if external Elasticsearch archetype was documented (introduced by @mattflax in #65) was documented in the wiki:
https://github.com/SeaseLtd/rated-ranking-evaluator/wiki/Maven%20Archetype

(I had to go through code/GH issues to find out about it)

mattflax:
I've updated the Archetypes wiki page to mention the external Elasticsearch archetype.

agazzarini:
LGTM!

PR #112, opened 2020-03-26
Add Maven plugin to connect to any search API

mattflax:
This PR adds a Maven plugin allowing any implementation of SearchPlatform to be used to call an external search API. I've also included an archetype, although it requires more changes than the Solr and ES archetypes to get working - there's a couple of README files with details of how to implement the connectors.


anthonygroves:
@mattflax A big +1 from me on this idea of connecting to a search API!
Question - how does this method handle versioning? I see in the `settings.json` for each version you can define the search API url. Does versioning/comparison require that you are able to query multiple variations of your search API at the same time?

mattflax:
@anthonygroves You have two options here:
1. have different versions of your API available and query them via different URLs;
2. there is a boolean option in the Persistence Configuration, `useTimestampAsVersion`, which uses the epoch milliseconds value at the start of the evaluation as the version. It **does** restrict you to a single configuration set, though. It is most likely to be useful when outputting results to an external store, such as Elasticsearch, but it applies with any persistence handler. (It was also undocumented, unfortunately - I've added a note about it to the [persistence wiki page](https://github.com/SeaseLtd/rated-ranking-evaluator/wiki/Persisting-evaluation-outputs).

agazzarini:
Hi @mattflax, great stuff, as usual.

mattflax:
Thanks @agazzarini - hope you're doing okay in these interesting times!

agazzarini:
You're welcome @mattflax. Yes, everything is ok, and I agree with you...interesting times! Hope the same there!

PR #113, opened 2020-03-26
Allow ratings at Query level, in addition to QueryGroup

mattflax:
We have a client who wishes to just specify queries, without Topic or QueryGroup blocks. However, the current behaviour requires ratings to be set at QueryGroup level, applying to all Query entries underneath - we required ratings to be set on individual Query entries. This PR adds that behaviour, allowing ratings to be specified at both Query and QueryGroup level.

- Ratings at Query level override any set at QueryGroup level (QueryGroup ratings are ignored);
- If a Query has no ratings, it will use ratings from the QueryGroup (as per current behaviour).

I hope this makes sense - let me know if it doesn't.

### DIFF
--- a/rre-core/src/main/java/io/sease/rre/core/Engine.java
+++ b/rre-core/src/main/java/io/sease/rre/core/Engine.java
@@ -255,7 +255,9 @@ public class Engine {
 
                                                     LOGGER.info("\t\tQUERY: " + queryString);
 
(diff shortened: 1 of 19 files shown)


agazzarini:
Absolutely @mattflax: as you already got , RRE has born from real use cases...I never met such scenario but I agree with you that can be useful

Issue #91, opened 2019-06-18
No errors shows in console when Solr cannot be reached


AlexAddisonLN:
When Solr is not reachable (e.g. RRE is run in a sandboxed environment) RRE will appear to run incredibly quickly, and inform us at the end that no evaluations were performed. It will also generate an invalid Excel file if Excel report generation is set up.

agazzarini:
Hi @AlexAddisonLN, many thanks for entering this. 
I assume you're talking about the external Solr (not the embedded one) right?

AlexAddisonLN:
Yes, I am talking about the external Solr

agazzarini:
I tried to reproduce your context; if Solr is not running I got this results: 

**External Solr Plugin**
```
[INFO ] *********************************
[INFO ] RRE: Index name => core1
[INFO ] RRE: ID Field name => id
[INFO ] Preparing platform for core1
[INFO ] RRE: Loading the Test Collection into External Apache Solr, configuration version v1.0
[INFO ] RRE: Loading the Test Collection into External Apache Solr, configuration version v1.1
[INFO ] RRE: External Apache Solr has been correctly loaded.
[INFO ] RRE: target versions are v1.0,v1.1
[INFO ] TOPIC: Fender basses
[INFO ] 	QUERY GROUP: The group tests several searches on the Fender brand
[INFO ] 		QUERY: fender
[INFO ] 		QUERY: fender Bass
[INFO ] 		QUERY: Fender
[INFO ] 	QUERY GROUP: Several searches on a given model (Jazz bass)
[INFO ] 		QUERY: jazz
[INFO ] 		QUERY: Jazz bass
[INFO ]   ... completed 0 / 5 evaluations ...
[INFO ] RRE: External Apache Solr Evaluation complete - preparing for shutdown
[WARN ] No queries recorded - returning empty evaluation
[INFO ] RRE: External Apache Solr Search Platform shutdown procedure executed.
[INFO ] RRE: Stopping persistence manager
```

The (Spreadsheet) Maven Report Plugin instead produces an empty excel. 
This behaviour is caused by the default evaluation manager which executes the queries in parallel and it doesn't block the evaluation if there's some error on a query. Unfortunately in the edge case when all queries fail (because Solr is not reachable) nothing is stopped and, as you said, the evaluation is incredibly quickly.

[...]

PR #126, opened 2020-11-01
Add check for platform availability before evaluating queries

mattflax:
This is an attempt to fix issue #91

It adds a platformCheck() method to the SearchPlatform API, with a check in the Engine evaluate loop to make sure all collections can be reached before starting the evaluation process. Messaging has been changed in the case that no queries have been evaluated, and warning messages appear stating that the collection/index isn't available.

I've included integration tests for the external platform modules, although I wasn't able to get the Solr integration test working through Zookeeper - I'm pretty sure it's an issue with Docker hostnames, but I don't think it can be easily fixed.


agazzarini (approved):
Great work as usual @mattoscln

Issue #128, opened 2020-11-02
Allow overriding the default Solr collection when using the external connector


mattflax:
Currently the external Solr connector only allows connecting to the collection specified in the ratings file. If we have changed our mappings/analysis and written to a new collection, it would be useful to test that collection alongside the previous version.

The external Solr archetype currently has a `collectionName` entry in the solr-settings.json file which is ignored. This feature request should allow that value to be different from the value in the ratings file, and for that collection to be used.

PR #130, opened 2020-11-23
#128: Implement collection overrides for external Solr:

mattflax:
This PR adds the ability to override the external Solr collection by specifying the alternative in the configuration file.

(The Maven archetype for the external Solr connector already had the collectionName property present in the config, but it was ignored. I'm not sure why the override wasn't implemented at the time.)

If merged, I'll modify the wiki pages as necessary.


agazzarini:
Thanks @mattflax looks great!

PR #149, opened 2025-07-07
Ticket/dage 10 project structure

Intrinsical-AI:
Added DAGE-10: added project structure with basic files, loggin, entrypoint and dummy tests


dantuzi:
Replaced by another equivalent PR

PR #150, opened 2025-07-07
DAGE-10: Added project structure

dantuzi:
### Notes
Adding the first project structure with logging and a simple dummy test

### Ticket
https://sease.atlassian.net/browse/DAGE-10


rre-dataset-generator/README.md
```diff
@@ -1,2 +1,104 @@
...
+
+### 1. Installation
+
+- [uv](https://github.com/astral-sh/uv): A fast Python package installer and resolver.
+- Python >=3.10
+
+First, create and activate a virtual environment using `uv`:
+
+```bash
+# Create the virtual environment in .venv
```
dantuzi: We should have a code block for each command so the README user can copy/paste the single command

rre-dataset-generator/README.md
```diff
@@ -1,2 +1,104 @@
...
+uv init
```
nicolo-rinaldi: This should be for the first initialisation of the repo. If the uv configuration files (e.g. uv.lock, pyproject.toml) are already in the repo, only `uv sync` is needed.

rre-dataset-generator/README.md
```diff
@@ -1,2 +1,104 @@
...
+outputDestination: "./output/" # Directory to save results
+outputExplanation: false # Whether the LLM should provide explanations
+```
+
+### 4. Running the Generator
+
+Execute the main script via the `typer` CLI, pointing to your configuration file:
+
+```bash
+python main.py run --arg1 --arg2 # example --config ./config.yaml --reasoning --save_reasoning
```
nicolo-rinaldi: replace `python` with `uv run` to use the uv environment.

rre-dataset-generator/README.md
```diff
@@ -1,2 +1,104 @@
...
+uv init
+
+# Venv generation
+uv venv
+
+# Install all dependencies from pyproject.toml
+uv pip install -e '.[dev,test]'
+
+# Activate the venv (Linux/macOS)
+source .venv/bin/activate
```
nicolo-rinaldi: No need to install any dependency or interact with the environment, just `uv sync` is doing everything.

rre-dataset-generator/README.md
```diff
@@ -1,2 +1,104 @@
...
+```
+
+
+### 2. Configuration
+
+Create a `config.yaml` file in the root directory. This file controls the entire generation process. E.g:
+
+```yaml
```
nicolo-rinaldi, line 23: TO BE CHANGED

rre-dataset-generator/README.md
```diff
@@ -1,2 +1,104 @@
...
+
+```bash
+# Create the virtual environment in .venv
+uv init
+
+# Venv generation
+uv venv
+
+# Install all dependencies from pyproject.toml
+uv pip install -e '.[dev,test]'
```
dantuzi: It doesn't work to another computer

dantuzi: Also, I see this tool `uv` creates a myproject.toml file with all dependencies to be downloaded.
@nicolo-rinaldi do you know where all those dependencies are found? I don't see the standard `requirements.txt` file

rre-dataset-generator/dataset_generator/__init__.py
dantuzi, line 1: I guess we don't need an empty file

rre-dataset-generator/README.md
```diff
@@ -1,2 +1,104 @@
...
+
+# LLM and Output
+relevanceScale: "binary" # or "graded"
+llmConfigurationFile: "llm_config.json" # Placeholder for LLM settings
+outputFormat: "Quepid" # or RRE, MTEB Leaderboard
+outputDestination: "./output/" # Directory to save results
+outputExplanation: false # Whether the LLM should provide explanations
+```
+
+### 4. Running the Generator
```
dantuzi: It should be n.3

rre-dataset-generator/README.md
```diff
@@ -1,2 +1,104 @@
...
+
+```bash
+python main.py run --arg1 --arg2 # example --config ./config.yaml --reasoning --save_reasoning
+```
+
+The script will then:
+1.  Fetch documents from the specified search engine.
+2.  Generate or load queries.
+3.  Score the relevance for each (document, query) pair.
+4.  (Future) Save the output to the specified destination.
```
dantuzi, line 75: We need a section that describes how to run the unit tests

rre-dataset-generator/tests/unit/test_dataset_generation_dummy.py
```diff
@@ -0,0 +1,32 @@
...
+sys.path.insert(0, project_root)
+
+from dataset_generator.main import app
+from dataset_generator.logger import configure_logging
+
+# Configure logging as soon as the module is loaded
+configure_logging()
+
+
+def test_argparse_yaml_with_list(caplog):
```
dantuzi: When you run the test, it fails with
```
ModuleNotFoundError: No module named 'dataset_generator.main'
```

rre-dataset-generator/tests/unit/test_dataset_generation_dummy.py
```diff
@@ -0,0 +1,32 @@
...
+    runner = CliRunner()
+    # Pass config_path as a positional argument
+    result = runner.invoke(app, ["src/yaml_configs/config.yml"])
+    
+    # Print the logs for debugging
+    for record in caplog.records:
+        print(f"LOG: {record.message}")
+    
+    # Check if the log message appears in any of the log records
+    assert any("Starting dataset generation process..." in record.message
```
dantuzi: What does it want to assert? Do we want to verify that all messages are "Starting dataset generation process..."?
BTW, at this stage, we simply want a dummy test because we want to validate that the tests are well configured

rre-dataset-generator/tests/unit/test_dataset_generation_dummy.py
```diff
@@ -0,0 +1,32 @@
...
+    result = runner.invoke(app, ["src/yaml_configs/config.yml"])
+    
+    # Print the logs for debugging
+    for record in caplog.records:
+        print(f"LOG: {record.message}")
+    
+    # Check if the log message appears in any of the log records
+    assert any("Starting dataset generation process..." in record.message 
+              for record in caplog.records)
+    assert result.exit_code == 0
```
dantuzi: It's always preferable to leave a blank line at the end of each file

rre-dataset-generator/dataset_generator/logger.py
```diff
@@ -0,0 +1,114 @@
...
+    "message", "msg", "name", "pathname", "process", "processName",
+    "relativeCreated", "stack_info", "thread", "threadName",
+}
+
+
+class JsonFormatter(logging.Formatter):
+    """
+    Formats log records as a single line of JSON.
+    
+    This formatter ensures that logs are machine-readable, which is ideal for
```
dantuzi: Why do we have logs formatted to be machine-readable? Our script is not supposed to be monitored by any of the log aggregator tools like ELK, Datadog, or Splunk. We want our logs to be human-readable

rre-dataset-generator/README.md
```diff
@@ -1,2 +1,104 @@
...
+├── pyproject.toml                      # Project metadata and dependencies
+├── queries.txt                         # Optional file with user-provided queries
+├── dataset_generator/                  # Core application package
+│   ├── __init__.py                     # Exposes the main DatasetGenerator service
+│   ├── cache.py                        # Caching utilities 
+│   ├── config.py                       # Pydantic model for parsing and validating config.yaml
+│   ├── logger.py                       # Structured logging configuration
+│   ├── models.py                       # Pydantic models for internal data structures
+│   ├── prompts.py                      # Prompts for the LLM 
+│   ├── llm/                            # Sub-package for Large Language Model services
```
dantuzi: A few folders don't exist

rre-dataset-generator/README.md
nicolo-rinaldi: Yes, the dependencies are built automatically during the installation of the python packages. With uv, the installation must be done using `uv add` instead of `pip install`. In this way the file is written automatically.

rre-dataset-generator/README.md
nicolo-rinaldi: I'll write it down just to remind myself when updating the readme: you just need to run the following
```
pytest <folder-with-test-files>
```

rre-dataset-generator/dataset_generator/__init__.py
nicolo-rinaldi: `__init__.py` files are used in python to distinguish between normal folders and packages one. I don't know if it is needed, but in a lot of repo is like that.

rre-dataset-generator/tests/unit/test_dataset_generation_dummy.py
nicolo-rinaldi: solved

rre-dataset-generator/tests/unit/test_dataset_generation_dummy.py
nicolo-rinaldi: solved

rre-dataset-generator/dataset_generation.py
nicolo-rinaldi, line 1: I changed the `typer` framework with the `argparse` one. This is a choice mainly due to the fact that, in our framework, we want to pass the config "dictionary" after parsing to a different set of functions. I could not find a way to do it using `typer`, so I changed framework.

rre-dataset-generator/tests/unit/test_dataset_generation_dummy.py
nicolo-rinaldi: Changed to just dummy tests

rre-dataset-generator/dataset_generator/llm/.gitkeep
nicolo-rinaldi, line 1: Replaced with a `__init__.py` file, to show the IDE that the llm folder is a package.

rre-dataset-generator/dataset_generator/search_engine/.gitkeep
nicolo-rinaldi, line 1: Replaced with a `__init__.py` file, to show the IDE that the llm folder is a package.

rre-dataset-generator/Makefile
nicolo-rinaldi, line 1: Why do we have this file?

nicolo-rinaldi (commented):
# Reviewed
- all files but `logger.py`

rre-dataset-generator/README.md
```diff
@@ -1,2 +1,109 @@
 # Dataset Generator
 
+This project provides a flexible command-line tool to generate relevance datasets for search evaluation. It can retrieve
+documents from a search engine, generate synthetic queries, and score the relevance of document-query pairs using LLMs.
+
+
+## Project Structure
+
+```bash
```
dantuzi: Not a blocker, but I'm afraid this section will be really difficult to maintain.
I would propose to remove it entirely or keep only the main folders.
For example, instead of keeping the folder "search_engine" with all the files, we can just keep the folder "search_engine"

rre-dataset-generator/dataset_generator.py
```diff
@@ -0,0 +1,29 @@
...
+"""
+
+import argparse
+from src.logger import configure_logging
+import logging
+
+def parse_args():
+    parser = argparse.ArgumentParser(description='Parse arguments for dataset_generator CLI.')
+
+    parser.add_argument('-d', '--dummy', type=str,
```
dantuzi, line 14: I guess we don't need it

rre-dataset-generator/pyproject.toml
```diff
@@ -0,0 +1,16 @@
...
+dependencies = [
+    "pydantic>=2.0",
+    "pytest>=8.4.1",
+    "pyyaml>=6.0.2",
+    "typer>=0.16.0",
+    "pytest>=7.4",
+]
+
+[project.scripts]
+dataset-generator = "main:app"
```
dantuzi: I saw that when I run the `uv sync` command with this configuration, it raises a Warning saying
```
warning: Skipping installation of entry points (`project.scripts`) because this project is not packaged; to install entry points, set `tool.uv.package = true` or define a `build-system`
```

rre-dataset-generator/src/logger.py
```diff
@@ -0,0 +1,114 @@
...
+
+    This function is idempotent and safe to call multiple times. It checks if
+    handlers are already configured to avoid duplication.
+    """
+    root_logger = logging.getLogger()
+    if not root_logger.hasHandlers():
+        root_logger.setLevel(level)
+        handler = logging.StreamHandler(sys.stdout)
+        handler.setFormatter(JsonFormatter())
+        root_logger.addHandler(handler)
```
dantuzi: How about something like that?
```
    logging.basicConfig(
        # filename='logs/log_file_name.log',
        level=logging.INFO,
        format='[%(asctime)s] %(filename)-15s:%(lineno)-4d %(levelname)-7s - %(message)s',
        datefmt='%H:%M:%S'
    )
```

The version I wrote with the commented filename goes to the stdout, otherwise it writes a log file into the folder `rre-dataset-generator/logs` (in that case, we need to create that logs folder to avoid exceptions.

dantuzi: Bear in mind that if you uncomment the filename, it works correctly only when you run the script from the `dataset_generator.py` entry point. If you want to test the code below, it will fail

rre-dataset-generator/tests/unit/test_dataset_generation_dummy.py
```diff
@@ -0,0 +1,40 @@
...
+#     result = runner.invoke(app, ["src/yaml_configs/config.yml"])
+#
+#     # Print the logs for debugging
+#     for record in caplog.records:
+#         print(f"LOG: {record.message}")
+#
+#     # Check if the log message appears in any of the log records
+#     assert any("Starting dataset generation process..." in record.message
+#               for record in caplog.records)
+#     assert result.exit_code == 0
```
dantuzi: I would avoid committing the commented code like that

rre-dataset-generator/README.md
```diff
@@ -1,2 +1,109 @@
...
+
+First thing to do is to check if the environment is active. If not, execute (in Unix based machines) the following 
+command to activate it:
+```bash
+source .venv/bin/activate
+```
+
+Now that the environment is active, execute `pytest` command as follows:
+```bash
+pytest tests/
```
dantuzi: from the bash it works only the command `pytest`

rre-dataset-generator/uv.lock
dantuzi, line 1: Do we need to commit this file?

rre-dataset-generator/.gitignore
```diff
@@ -0,0 +1,14 @@
...
+dist/
+wheels/
+*.egg-info
+
+# Virtual environments
+.venv
+
+# Local Files and Directories
+.output_dataset
+output/
```
dantuzi, line 15: Don't we need to add the `uv.lock` file?

rre-dataset-generator/dataset_generator.py
nicolo-rinaldi: This is a temporary file to do a dummy test.

nicolo-rinaldi (approved):

PR #152, opened 2025-07-09
DAGE-4: Added code to deal with config file

nicolo-rinaldi:
## Notes
Added code to deal with config file (parsing, testing, CLI interface). Code added:
- config.py
- test_config.py
- all files needed to test an initial config
- added logging to 'config.py' and 'dataset_generator.py' files

## Ticket
https://sease.atlassian.net/jira/software/projects/DAGE/boards/84?selectedIssue=DAGE-4


rre-dataset-generator/config.yaml
```diff
@@ -0,0 +1,64 @@
...
+
+# Maximum number of documents to retrieve from the search engine to generate queries
+docNumber: 100
+
+# List of fields from documents used to generate context and relevance scoring
+# These should match fields available in the search engine schema
+# Must be at least one field
+docFields:
+  - "title"
+  - "body"
```
nseidan: we have "description" field instead of "body"

rre-dataset-generator/dataset_generator.py
```diff
@@ -1,29 +1,27 @@
...
     args = parse_args()
 
     configure_logging(level=logging.DEBUG)
     log = logging.getLogger(__name__)
 
     try:
-        log.debug(f"Dummy variable {args.dummy} loaded successfully.")
+        config = load_config(args.config_file)
+        log.debug("Configuration loaded successfully.")
+        # print(config.model_dump_json(indent=2))
```
nseidan: redundant # print(config.model_dump_json(indent=2)) ?

rre-dataset-generator/src/config.py
```diff
@@ -0,0 +1,71 @@
+from typing import List, Optional, Literal, Dict
+from pydantic import BaseModel, HttpUrl, Field, field_validator, FilePath
+import yaml
+from pathlib import Path
+
+
+class Config(BaseModel):
+    QueryTemplate: Optional[str] = Field("q=#$query##", description="Template string for queries with a placeholder for keywords.")
```
nseidan: naming convention: better to use  lowercase with underscores for field names e.g. query_template, search_engine_type ("solr", "vespa" ...) ?

rre-dataset-generator/src/config.py
```diff
@@ -0,0 +1,71 @@
...
+    def check_doc_type(cls, v):
+        if v is not None:
+            if v.suffix[1:] != "txt":
+                raise ValueError("queries' file must have TXT extension")
+        return v
+
+    @field_validator('LLMConfigurationFile')
+    def check_config_type(cls, v):
+        if v is not None:
+            if v.suffix[1:] != "yaml" and v.suffix[1:] != "yml":
```
nseidan: more readable:  `if v.suffix[1:] not in {"yaml", "yml"}:`

rre-dataset-generator/src/config.py
dantuzi: +1
I would follow this standard: https://google.github.io/styleguide/pyguide.html#316-naming

rre-dataset-generator/config.yaml
```diff
@@ -0,0 +1,64 @@
+# (Optional) Template string for generated/given queries used to search documents and build relevance
+# judgemenets, with a placeholder for inserting keywords
+# Example (Solr):
```
dantuzi: This is not the query used to search documents and build relevance judgements.

This is the query the user wants to evaluate, and to do that, we are generating the entire dataset.

rre-dataset-generator/config.yaml
```diff
@@ -0,0 +1,64 @@
...
+query_template: "q=#$query##&fq=genre:horror&wt=json"
+
+# The type of search engine to use
+# Accepted values: Solr, Elasticsearch, Opensearch, Vespa
+search_engine_type: "solr"
```
dantuzi: Are the accepted values uppercase or lowercase?
The documentation says we accept "Solr" but the example value is "solr".
Is it case-insensitive?

rre-dataset-generator/src/config.py
```diff
@@ -5,25 +5,25 @@
...
-    RelevanceScale: Literal['Binary', 'Graded']
-    LLMConfigurationFile: FilePath = Field(..., description="Path to the LLM configuration file.")
-    OutputFormat: Literal['Quepid', 'RRE']
-    OutputDestination: Path = Field(..., description="Path to save the output dataset.")
-    OutputExplanation: Optional[bool] = Field(False, description="Whether to generate an explanation file.")
+    generate_queries_from_documents: Optional[bool] = True
+    total_num_queries_to_generate: int = Field(..., gt=0, description="Total number of queries to generate.")
+    relevance_scale: Literal['binary', 'graded']
+    llm_configuration_file: FilePath = Field(..., description="Path to the LLM configuration file.")
+    output_format: Literal['Quepid', 'RRE']
```
nseidan: @nicolo-rinaldi minor, to lowercase -> quepid, rre ? to keep it consistent with others like binary/graded etc

rre-dataset-generator/tests/unit/resources/invalid_type.yaml
```diff
@@ -1,9 +1,9 @@
...
-OutputDestination: "out.json"
+search_engine_type: "solr"
+search_engine_collection_endpoint: "http://localhost"
+doc_number: "one hundred"  # Invalid type
+doc_fields: ["title"]
+total_num_queries_to_generate: 5
+relevance_scale: "binary"
+llm_configuration_file: "llm_config.yaml"
+output_format: "RRE"
+output_destination: "out.json"
```
nseidan: @nicolo-rinaldi rename to output.json ? more meaningful.

rre-dataset-generator/config.yaml
```diff
@@ -0,0 +1,64 @@
...
+
+# Endpoint URL of the search engine collection or index
+search_engine_collection_endpoint: "http://localhost:8983/solr/mycore"
+
+# (Optional) Filter query to restrict the set of documents used to generate queries. Top k retrieved documents
+# are used, with k computed accordingly looking at the params: docNumber, totalNumQueriesToGenerate
```
dantuzi: When we are retrieving documents, we only use `docNumber`.
Maybe I understand why you also add the parameter `totalNumQueriesToGenerate`, you want to say that if the number of provided queries is higher than the `totalNumQueriesToGenerate`, we don't need to extract any document from Solr. In that case, I would avoid this at this stage because it can be confusing.
I wouldn't say anything about getting the "top k" documents, but if you want to say that, I would only mention the docNumber.

nseidan (approved):

rre-dataset-generator/config.yaml
```diff
@@ -0,0 +1,64 @@
...
+# (Optional) File containing predefined queries to use instead of or alongside generated ones
+# If available, must be in a JSON format.
+queries: "queries.txt"
+
+# (Optional) Whether to generate queries from documents
+# Default: true; set to false to disable query generation from documents
+generate_queries_from_documents: true
+
+# Total number of queries to generate (includes predefined queries, if any)
+total_num_queries_to_generate: 10
```
dantuzi: I'm sorry for mentioning it now, but I believe this name can be misleading.
If I read "total number of queries to generate" I'm expecting the software always generates that number of queries (but it can be less if there are enough provided queries).
How about something more generic like `num_queries` or `num_queries_needed` (any better idea is welcome)

rre-dataset-generator/config.yaml
```diff
@@ -0,0 +1,64 @@
...
+# Output format for the generated dataset
+# Accepted values: Quepid, RRE
+output_format: "Quepid"
+
+# Path where the output dataset will be saved
+output_destination: "output/generated_dataset.json"
+
+# Whether to generate an additional file with query, document, score, and explanation
+# Accepted values: true, false
+output_explanation: true
```
dantuzi: I remember we discussed having two parameters instead of just one to be added for version V2. [Notes](https://docs.google.com/document/d/1oCofhKdphXUJgQJOvLZ-NLeVvirLMm84JATXJR8jAOI/edit?tab=t.0)
```
Add parameters for reasoning (for the LLM) and save_explanation (save the explanation file) if save_explanation=true => reasoning=true
```

That being said, I'd remove this parameter for now, and we can add it as part of the V2 implementation task

rre-dataset-generator/llm_config.yaml
dantuzi, line 1: Why do we need to commit this empty file?

rre-dataset-generator/queries.txt
dantuzi, line 1: Same as before. We shouldn't commit this file because the user can link any file

rre-dataset-generator/pyproject.toml
```diff
@@ -5,6 +5,7 @@ description = "Dataset Generator project"
 readme = "README.md"
 requires-python = ">=3.10"
 dependencies = [
+    "argparse>=1.4.0",
```
dantuzi, line 8: Did we forget to add this in the previous PR?

rre-dataset-generator/src/config.py
```diff
@@ -0,0 +1,71 @@
...
+
+    :param config_path: Path to the YAML config file
+    :return: Parsed and validated Config object
+    """
+    # try:
+    with open(config_path, 'r') as f:
+        raw_config = yaml.safe_load(f)
+    return Config(**raw_config)
+    # except Exception as e:
+    #     raise RuntimeError(f"Error loading config: {e}")
```
dantuzi: Commented lines like these are not needed.
I think an exception with a descriptive message may be useful. Is it already provided by `safe_load` function?

rre-dataset-generator/src/config.py
```diff
@@ -0,0 +1,71 @@
...
+# Example usage in pipeline
+if __name__ == "__main__":
+    from src.logger import configure_logging
+    import logging
+
+    configure_logging(level=logging.DEBUG)
+    log = logging.getLogger(__name__)
+
+    config = load_config("config.yaml")
+    log.debug("Configuration loaded successfully.")
```
dantuzi: Why this? Sorry for my ignorance, but isn't it possible to test this using the unit tests?

rre-dataset-generator/tests/unit/test_config.py
```diff
@@ -0,0 +1,82 @@
...
+    assert config.llm_configuration_file == FilePath("llm_config.yaml")
+
+def test_output_format(config):
+    assert config.output_format == "Quepid"
+
+def test_output_destination(config):
+    assert config.output_destination == Path("output/generated_dataset.json")
+
+def test_output_explanation(config):
+    assert config.output_explanation is True
```
dantuzi: All of those tests can be written in a single test case.

```
def test_good_config_expect_all_parameters_read:
    config = load_config(good file)
    assert parameter 1
    assert parameter 2
    ...
    assert parameter n
```

rre-dataset-generator/tests/unit/test_dataset_generation_dummy.py
```diff
@@ -13,4 +11,4 @@ def args():
     return parse_args()
 
 def test_argparse_with_defaults(args):
-    assert args.dummy == "default"
\ No newline at end of file
+    assert args.config_file == "config.yaml"
```
dantuzi: Can you please add a new line at the end?

dantuzi (changes requested):

rre-dataset-generator/config.yaml
nicolo-rinaldi: I forgot to change it. I did it now

rre-dataset-generator/config.yaml
nicolo-rinaldi: I changed it in `num_queries_needed`

rre-dataset-generator/src/config.py
nicolo-rinaldi: Yes, I commented this code to have more descriptive exception. I removed now also the comments

rre-dataset-generator/src/config.py
nicolo-rinaldi: Yes, it is. I removed this code

rre-dataset-generator/tests/unit/test_config.py
nicolo-rinaldi: Done

rre-dataset-generator/llm_config.yaml
nicolo-rinaldi: I used it to run test, since pydantic datatype does automatic validation and needs an actual file. Maybe I can move it to the `tests/unit/resources/`?

rre-dataset-generator/queries.txt
nicolo-rinaldi: Same as last comment

rre-dataset-generator/pyproject.toml
nicolo-rinaldi: It should be in python base library, but it was raising me some errors, so I updated the library `argparse`

rre-dataset-generator/tests/unit/test_dataset_generation_dummy.py
nicolo-rinaldi: done

rre-dataset-generator/config.yaml
nicolo-rinaldi: Ok, an idea can be also to put this param inside the `llm_config` file. I don't know if it make sense since is an option that reflects also in the structure of the output dataset

nicolo-rinaldi: Anyway, for now i removed the `output_explanation` param

rre-dataset-generator/config.yaml
nicolo-rinaldi: Changed description. I did not understood fully the use of this param. Now should be fine

rre-dataset-generator/config.yaml
nicolo-rinaldi: removed the mention to top k retrieved docs

rre-dataset-generator/llm_config.yaml
nicolo-rinaldi: I decided to move the file in the mentioned folder

rre-dataset-generator/pyproject.toml
dantuzi: Just out of curiosity, what error did you get?

rre-dataset-generator/tests/unit/test_dataset_generation_dummy.py
```diff
@@ -1,5 +1,3 @@
-import sys
-import os
 from src.logger import configure_logging
```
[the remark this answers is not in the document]
dantuzi: it is used

PR #154, opened 2025-07-10
DAGE-6: Add  LLM Service with Langchain Integration

nseidan:
* Added LLM service class,` llm_config.yaml ` and `llm_config.py` files for the LLM configuration 
* In `llm_provider_factory.py`  for the sake of simplicity added only two LLM providers support: `openai` and `gemini`
* Added unit test

Run manual test to verify LLMService.test_connection() returns some response:
export `OPENAI_API_KEY` via .env file and got the following response: 
```
LLM answer: Sure! Here are greetings in five different languages:

1. English: Hello
2. Spanish: Hola
3. French: Bonjour
4. German: Hallo
5. Japanese: こんにちは (Konnichiwa)

```


rre-dataset-generator/manual_test.py
```diff
@@ -0,0 +1,27 @@
+from src.llm.factory import build_chat_model
+from src.llm.llm_config import LLMConfig
+from src.llm.llm_service import LLMService
+
+
+def main():
+    # 1. Make sure LLM API key is set: e.g. export OPENAI_API_KEY=sk-...
```
nicolo-rinaldi: another option (instead of doing the export each time we open the terminal) can be to define the file `.env` and put it in the `.gitignore`. In this case, we need also to update the `README.md` file to tell the user how to set up the environment variable, but it's easier for multiple usage. 
(just a possiblity)

rre-dataset-generator/src/llm/llm_service.py
```diff
@@ -0,0 +1,15 @@
...
+    def __init__(self, chat_model: BaseChatModel):
+        self.chat_model = chat_model
+
+    def test_connection(self, prompt: str = "Say hello in five different languages") -> str:
+        """
+        Send a test prompt to verify the model returns a response.
+        """
+        response = self.chat_model.invoke([HumanMessage(content=prompt)])
+        return response.content.strip()
+
```
nicolo-rinaldi, line 21: I would add another class with a `message` variable more structured, for example:
```python
messages = [
    SystemMessage(
        content="You are a helpful assistant! Your name is Bob."
    ),
    HumanMessage(
        content="What is your name?"
    )
]
```
[from [LangChain documentation](https://api.python.langchain.com/en/latest/messages/langchain_core.messages.human.HumanMessage.html#langchain_core.messages.human.HumanMessage)]

rre-dataset-generator/src/llm/llm_config.py
```diff
@@ -0,0 +1,32 @@
...
+    max_tokens: int = Field(default=512, gt=0)
+    api_key_env: Optional[str] = None
+
+
+class LLMConfig(BaseModel):
+    default_provider: str
+    providers: Dict[str, ProviderCfg]
+
+    @model_validator(mode="after")
+    def validate_default_provider_exists(self) -> "LLMConfig":
```
nicolo-rinaldi: why are there the double quotes in the return type?

nicolo-rinaldi (approved):

rre-dataset-generator/llm_config.yaml
```diff
@@ -0,0 +1,28 @@
...
+
+providers:
+  openai:
+    # OpenAI chat model name
+    model: gpt-4o
+
+    # Controls randomness of the model output:
+    #  0.0 = completely deterministic
+    #  1.0 = maximal diversity
+    temperature: 0.3
```
dantuzi: Is it possible to extract this as a constant and apply it to both models? (same for the max token)
Or maybe those parameters strictly depend on the model and they can be different

rre-dataset-generator/manual_test.py
dantuzi, line 1: Why do we need a manual test file?

rre-dataset-generator/tests/unit/test_llm_service.py
```diff
@@ -0,0 +1,12 @@
...
+from src.llm.llm_service import LLMService
+
+
+def test_llm_service_test_connection():
+    fake_llm = FakeListChatModel(responses=["Hello, world!"])
+    service = LLMService(chat_model=fake_llm)
+
+    response = service.test_connection()
+    assert isinstance(response, str)
+    assert "world" in response
```
dantuzi: Can't you test the response is "Hello, world!"?

rre-dataset-generator/src/llm/llm_config.py
dantuzi: +1

rre-dataset-generator/src/llm/llm_config.py
```diff
@@ -0,0 +1,32 @@
...
+        if self.default_provider not in self.providers:
```
dantuzi: Can you clarify the purpose of the "default_provider" configuration variable?
Probably I don't understand well how langchain has to be configured, but it looks like you are asking the client to configure more than one LLM, and the client has to choose what to use, am I correct?
In that case, the name "default" is a bit misleading. When I read the word "default" I'm expecting to use that value if the user doesn't define anything, which is not the case

rre-dataset-generator/manual_test.py
nseidan: Actually we don't need it. It is just for myself to verify the LLM is working and returning a result. I removed it.

rre-dataset-generator/tests/unit/test_llm_service.py
nseidan: yes we can; updated.

rre-dataset-generator/src/llm/llm_config.py
nseidan: In the beginning, I got NameError,  Pydantic seems to look for the class immediately even the class is not yet created.  In this case, LLMConfig, so wrapping into "", the Python tells the class is not yet created. 

Anyways, I explored another clean way to do, so then I removed the quotes, instead imported `from __future__ import annotations` This is available python 3.7+.

rre-dataset-generator/src/llm/llm_config.py
nseidan: @dantuzi well, default is actually the primary/active one. I would have named it active/currently being used. User has flexibility in the LLM config. One day they decide to use OpenAI, a month later, they might want to try Gemini, that's why, we have providers map to switch between LLMs.  

Alternatively, to make it easy, we can just have only one LLM provider in the llm config file. No primary/no providers map.  What do you think ?

rre-dataset-generator/manual_test.py
nseidan: makes sense; it is just for dev manual testing ; i shouldnt push this file.

rre-dataset-generator/src/llm/llm_service.py
nseidan: updated

rre-dataset-generator/llm_config.yaml
nseidan: those params are common for all LLM configs.

rre-dataset-generator/manual_test.py
nseidan: added;

nseidan:
@dantuzi  In the llm_config file, I kept only one provider, as user provides it. But in the` llm_provider_factory.py,` we have the support for openai and gemini and also in the future we can add more LLM providers including local LLM there, it is extendable. 


While looking into local LLM configuration,  it is feasible to extend` llm_provider_factory.py `file to introduce a method build_local() and in the llm_config file, we just provide name as `local` so that the system knows to load from the local directory path.

rre-dataset-generator/tests/unit/test_llm_service.py
```diff
@@ -0,0 +1,12 @@
+from langchain_core.language_models.fake_chat_models import FakeListChatModel
+from src.llm.llm_service import LLMService
+
+
+def test_llm_service_test_connection():
```
dantuzi, line 5: I tried to run this test and I got this failure

```
________________________ test_llm_service_test_connection ________________________

    def test_llm_service_test_connection():
        fake_llm = FakeListChatModel(responses=["Hello, world!"])
        service = LLMService(chat_model=fake_llm)
    
>       response = service.test_connection()
                   ^^^^^^^^^^^^^^^^^^^^^^^^^
E       TypeError: LLMService.test_connection() missing 1 required positional argument: 'prompt'

tests/unit/test_llm_service.py:9: TypeError
```

dantuzi (changes requested):

rre-dataset-generator/tests/unit/test_llm_service.py
nseidan: updated;

dantuzi (approved):

rre-dataset-generator/src/llm/llm_config.py
```diff
@@ -0,0 +1,25 @@
...
+from typing import Optional
+
+import yaml
+from pydantic import BaseModel, Field
+
+
+class LLMConfig(BaseModel):
+    name: str
+    model: str
+    temperature: float = Field(default=0.3, ge=0.0, le=1.0)
```
Intrinsical-AI, line 13: I think this is not correct. Temperature could be higher than 1. For example: OpenAI has a range of [0-2], with a default of 1. Anw: if the temperature it's intrinsic of each model;

Provider | Temp. Range
OpenAI Chat Completions |	0 – 2	
Anthropic Claude (v2 / v3)	| 0 – 1	(if > 1 -> Validation error)
Google Gemini |	0 – 2 (recomended 0-1)	
Google AI for Developers | any positive float. (higher = more random)

Also; why we would want to default temp=0.3 ?? Seems a random number; and as said, each provider work with it's ranges. We should allow LLMConfig.temperature to be None, and trust the default provider value

rre-dataset-generator/src/llm/llm_config.py
```diff
@@ -0,0 +1,25 @@
...
+    max_tokens: int = Field(default=512, gt=0)
```
Intrinsical-AI, line 14: . I'd allow LLMConfig.max_tokens to be None, and trust the default provider value. The idea for our usecase should be overriding if needed.

rre-dataset-generator/src/llm/llm_config.py
```diff
@@ -0,0 +1,25 @@
...
+    api_key_env: Optional[str] = None
+
+    @staticmethod
```
Intrinsical-AI, line 17: staticmethod makes a lot of sense, but we can compare it to classmethod. Because? What if tomorrow we have another configuration (e.g. AdvancedLLMConfig) that inherits from LLMConfig? -> @staticmethod will continue calling the LLMConfig.load method; that breaks the inheritance

I would change this for a class method, for long-term flexibility and good practices.

With the class method, cls would now be a load parameter, so when we call load, it will work on the current class, regardless of whether it is LLMConfig or a subclass of it

rre-dataset-generator/src/llm/llm_provider_factory.py
```diff
@@ -0,0 +1,54 @@
...
+
+from dotenv import load_dotenv
+from langchain_core.language_models import BaseChatModel
+from langchain_openai import ChatOpenAI
+from langchain_google_genai import ChatGoogleGenerativeAI
+from .llm_config import LLMConfig
+
+
+def build_openai(config: LLMConfig) -> BaseChatModel:
+    load_dotenv()  # load .env file
```
Intrinsical-AI, line 19: Id remove the comment, the name it's self-descriptive enough?

rre-dataset-generator/src/llm/llm_provider_factory.py
```diff
@@ -0,0 +1,54 @@
...
+PROVIDER_REGISTRY = {
+    "openai": build_openai,
+    "gemini": build_gemini,
+}
+
+
+def build_chat_model(config: LLMConfig) -> BaseChatModel:
+    provider_name = config.name
+    if provider_name not in PROVIDER_REGISTRY:
+        raise ValueError(f"Unsupported provider: {provider_name}")
```
Intrinsical-AI, line 53: add log.error / log.debug?

rre-dataset-generator/src/llm/llm_provider_factory.py
```diff
@@ -0,0 +1,54 @@
...
+        max_tokens=config.max_tokens,
+        api_key=key,
+    )
+
+
+def build_gemini(config: LLMConfig) -> BaseChatModel:
+    load_dotenv()  # load .env file
+    key = os.getenv(config.api_key_env or "GOOGLE_API_KEY")
+    if not key:
+        raise ValueError("Google API key not set.")
```
Intrinsical-AI, line 35: log.error / log.debug?

Intrinsical-AI (commented):
Critical: 

LLMConfig temperature -> Validations and default

Long-term improvements -> LLMConfig.load->  switch decorator; static to classmethod

Readability: log.error / log.debug instead value error with strings

rre-dataset-generator/src/llm/llm_config.py
nseidan: @Intrinsical-AI this is what i configured, not about it is correct or not; assuming that smaller the value is we keep the context ouput bind to the prompt. 
I agree with the default values what the model provides, we should leave it.

rre-dataset-generator/src/llm/llm_config.py
nseidan: well,  we should control the costs based on the output result size. I wouldn't change it, we should provide the default value otherwise, we might hit high costs.

rre-dataset-generator/src/llm/llm_provider_factory.py
nseidan: nope. it is not really,  making it visible so that, shouldnt forget to load it.

PR #155, opened 2025-07-14
DAGE-3: Implementation of search engine abstraction and class that implements it in Solr

nicolo-rinaldi:
## Changes
- Interface at `rre-dataset-generator/src/search_engine/interface.py`
- Code to interact with Solr search engine at `rre-dataset-generator/src/search_engine/solr_search_engine.py`
- Added tests in the `tests/integration/` folder with mocks (to see it they make sense done like that)

[Jira ticket](https://sease.atlassian.net/browse/DAGE-3)


rre-dataset-generator/dataset_generator.py
```diff
@@ -20,7 +11,7 @@ def parse_args():
     log = logging.getLogger(__name__)
 
     try:
-        config = load_config(args.config_file)
+        config = Config.load(args.config_file)
         log.debug("Configuration loaded successfully.")
     except Exception as e:
         log.debug(f"Error loading configuration: {e}")
```
nseidan: why log.debug ? not log.error if exception occurs ?

rre-dataset-generator/result.json
```diff
@@ -0,0 +1 @@
+[]
```
nseidan: There is no newline at the end, you can see git shows with the red circle

rre-dataset-generator/src/search_engine/solr_search_engine.py
```diff
@@ -0,0 +1,85 @@
...
+
+    def search(self, payload: Dict[str, Any]) -> List[Dict[str, Any]]:
+        """Search for documents using a query."""
+        search_url = os.path.join(str(self.endpoint), "select")
+
+        response = requests.post(search_url, headers=self.HEADERS, json=payload)
+        if response.status_code == 200:
+            return response.json()['response']['docs']
+        else:
+            raise ValueError()
```
nseidan: ValueError is very vague, better to throw an error with http code, if it is server error or client and also log error?

rre-dataset-generator/tests/integration/resources/good_config.yaml
```diff
@@ -0,0 +1,17 @@
+query_template: "q=description:#$query##"
+search_engine_type: "solr"
+search_engine_collection_endpoint: "http://localhost:8983/solr/testcore"
+documents_filter:
+  - title:
+      - "and"
```
nseidan, line 6: title is "and"?

rre-dataset-generator/src/search_engine/solr_search_engine.py
```diff
@@ -0,0 +1,85 @@
...
+
+    def extract_documents_to_evaluate_system(self, query_template: str, keyword: str="*:*") -> List[Dict[str, Any]]:
+        """Search for documents using a query."""
+        template = query_template.replace(self.PLACEHOLDER, keyword)
+        payload = self.template_to_json_body(template)
+        return self.search(payload)
+
+    def search(self, payload: Dict[str, Any]) -> List[Dict[str, Any]]:
+        """Search for documents using a query."""
+        search_url = os.path.join(str(self.endpoint), "select")
```
nseidan: I just looked up `os.path.join` and it seems  it is used to construct file system path, for Windows , it uses ' \ ' backslash instead of '/', so you might end up getting something like this ` ../testcore\select`  which might lead to http not found url.

rre-dataset-generator/src/search_engine/solr_search_engine.py
```diff
@@ -0,0 +1,85 @@
...
+                    if len(values) == 1:
+                        clause = f'{field}:{values[0]}'
+                    else:
+                        or_values = ' OR '.join(f'{v}' for v in values)
+                        clause = f'{field}:({or_values})'
+                    payload['params']['fq'].append(clause)
+
+        return self.search(payload)
+
+    def extract_documents_to_evaluate_system(self, query_template: str, keyword: str="*:*") -> List[Dict[str, Any]]:
```
nseidan: very long method name, maybe` fetch_for_evaluation`?

rre-dataset-generator/src/search_engine/solr_search_engine.py
```diff
@@ -0,0 +1,85 @@
...
+        for key, default_value in defaults.items():
+            if key not in json_body or not json_body[key]:
+                json_body[key] = [default_value]
+
+        return {
+            'query': json_body.get('q')[0],
+            'params': {k: v[0] for k, v in json_body.items() if k != 'q'}
+        }
+
+    def extract_documents_to_generate_queries(self,
```
nseidan: lengthy method name, maybe `fetch_for_query_generation` ?

rre-dataset-generator/src/search_engine/solr_search_engine.py
```diff
@@ -0,0 +1,85 @@
...
+        Converts a Solr query string into a structured JSON body.
+
+        Args:
+            template_payload (str): The Solr query string, e.g., 'q=ghosts&fq=genre:horror&wt=json'.
+
+        Returns:
+            dict: A dictionary representing the query parameters.
+        """
+        # Parse the query string into a dictionary
+        json_body = parse_qs(template_payload)
```
nseidan, line 37: does this method handle quotes and a question mark?

rre-dataset-generator/src/search_engine/solr_search_engine.py
```diff
@@ -0,0 +1,85 @@
...
+    """
+    Solr implementation to search into a given collection
+    """
+    def __init__(self, endpoint: str):
+        super().__init__(endpoint)
+        self.HEADERS = {'Content-Type': 'application/json'}
+
+    @staticmethod
+    def template_to_json_body(template_payload: str) -> dict:
+        """
```
nseidan, line 27: add return type as well like Dict[str, Any] to be consistent in the code.

rre-dataset-generator/src/search_engine/solr_search_engine.py
nseidan: how about adding unit test for `template_to_json_body` if it correctly handles edge cases, including documents_filter?

rre-dataset-generator/try_solr.py
```diff
@@ -0,0 +1,26 @@
...
+    config = Config.load(args.config_file)
+
+    search_engine = SolrSearchEngine('http://localhost:8983/solr/testcore/')
+
+    # docs = search_engine.extract_documents_to_generate_queries(documents_filter=config.documents_filter,
+    #                                                            doc_number=config.doc_number)
+
+    docs = search_engine.extract_documents_to_evaluate_system(keyword="and", query_template=config.query_template)
+
+    with open("result.json", "w") as file:
```
nseidan: adding solr to file name? solr_extracted_dataset or anything else ?

rre-dataset-generator/dataset_generator.py
nicolo-rinaldi: yeah, it makes sense. Fixed

rre-dataset-generator/tests/integration/resources/good_config.yaml
nicolo-rinaldi: yes, this is just to look if we have returned files

rre-dataset-generator/src/search_engine/solr_search_engine.py
nicolo-rinaldi: I hope the user working on Windows will introduce as a param in a the config file a viable Windows-like path. Am I wrong?

rre-dataset-generator/src/search_engine/solr_search_engine.py
nicolo-rinaldi: changed

rre-dataset-generator/try_solr.py
nicolo-rinaldi: Picked `solr_extracted_dataset.json` as name and added empty line at the end of the file

rre-dataset-generator/src/search_engine/solr_search_engine.py
nseidan: @nicolo-rinaldi I meant `os.path.join` is for file path construction, for http path, we should avoid using the os.path.join Also we shouldnt assume the user introduces its own param for windows os.

nseidan: @nicolo-rinaldi,  take a look into `urljoin`  https://stackoverflow.com/questions/10893374/python-confusions-with-urljoin, it seems works well for constructing urls.

rre-dataset-generator/src/search_engine/solr_search_engine.py
nicolo-rinaldi: I'll add more error handling (to review) in the next commit

rre-dataset-generator/result.json
nicolo-rinaldi: fixed

rre-dataset-generator/src/search_engine/solr_search_engine.py
nicolo-rinaldi: Fixed

rre-dataset-generator/pyproject.toml
Intrinsical-AI, line 1: I think argparse it's included by default in python std. lib for Python>=2.7 (therefore, in python 3.10). We can remove it from the toml file

rre-dataset-generator/src/search_engine/solr_search_engine.py
nicolo-rinaldi: yes, that's for sure a safer option. Fixed

rre-dataset-generator/dataset_generator.py
```diff
@@ -20,7 +11,7 @@ def parse_args():
     log = logging.getLogger(__name__)
 
     try:
-        config = load_config(args.config_file)
+        config = Config.load(args.config_file)
```
dantuzi: I like it

rre-dataset-generator/src/config.py
```diff
@@ -42,14 +44,14 @@ def check_config_type(cls, v):
...
-    Load and validate configuration from a YAML file.
-
-    :param config_path: Path to the YAML config file
-    :return: Parsed and validated Config object
-    """
-    with open(config_path, 'r') as f:
-        raw_config = yaml.safe_load(f)
-    return Config(**raw_config)
\ No newline at end of file
+    @staticmethod
```
Intrinsical-AI: staticmethod makes a lot of sense, but, we can compare with classmethod. Why? What if tomorrow we have another config (ej: AWSConfig) that inherits from Config? -> @staticmethod will still be calling the Config.load method; which breaks heritage; if we want to AWSConfig.load().

So I'd change this for a classmethod, for long term flexibility and good practices

Intrinsical-AI: Wth class method, cls it's now a parameter of load, so when we call load, it's gonna work on the current class, no matter if it's Config, or a subclass.

rre-dataset-generator/try_solr.py
dantuzi, line 1: Do we need to commit this file?
If we want to commit it, it should be somewhere in the integration test folder as a temporary test before we write the proper integration tests.
Furthermore, the file `result.json` should also be moved to be a temporary integration test

rre-dataset-generator/src/search_engine/solr_search_engine.py
Intrinsical-AI: I'd better add log.error and keeping ValueError() or some try-except, than defining custom Exceptions.

rre-dataset-generator/src/search_engine/interface.py
```diff
@@ -0,0 +1,30 @@
+from abc import ABC, abstractmethod
+from typing import List, Dict, Any, Union
+from pydantic import HttpUrl, BaseModel
+
+class EndpointValidator(BaseModel):
+    endpoint: HttpUrl
+
+class BaseSearchEngine(ABC):
+    def __init__(self, endpoint: str):
+        validated = EndpointValidator(endpoint=endpoint)
```
dantuzi: Does the HttpUrl type already validate the endpoint? How about if the validation fails?

rre-dataset-generator/dataset_generator.py
```diff
@@ -1,14 +1,5 @@
...
-    parser.add_argument('-c', '--config_file', type=str,
-                        help='Config file path to use for the application [default: \"config.yaml\"]',
-                        required=False, default="config.yaml")
-
-    return parser.parse_args()
+from src.config import Config
+from src.utils import parse_args
 
 if __name__ == "__main__":
     from src.logger import configure_logging
```
Intrinsical-AI: This doesn't make a lot of sense to me. The resource saving it's minimal.
I'd unify the logs / exception handling in all the classes, at the begining. Not in the main


**_from src.logger import configure_logging
import logging
log = logging.getLogger(__name__)_**

Intrinsical-AI (commented):
Small proposals

rre-dataset-generator/pyproject.toml
nicolo-rinaldi: I needed to update the dependencies and it came out. I think there are no problem if we keep the argparse dependency

rre-dataset-generator/src/config.py
nicolo-rinaldi: done

rre-dataset-generator/dataset_generator.py

rre-dataset-generator/src/search_engine/solr_search_engine.py
nicolo-rinaldi: done

rre-dataset-generator/src/search_engine/solr_search_engine.py
nicolo-rinaldi: I added tests and seems working as expected

rre-dataset-generator/src/config.py
```diff
@@ -25,31 +33,36 @@ class Config(BaseModel):
...
         if any(not f.strip() for f in v):
+            log.error("docFields cannot contain empty strings.")
             raise ValueError("docFields cannot contain empty strings.")
         return v
 
     @field_validator('queries')
     def check_doc_type(cls, v):
         if v is not None:
             if v.suffix[1:] != "txt":
+                log.error("queries' file must have TXT extension")
```
nseidan: here .txt extension right ? can you just replace TXT with .txt ?

rre-dataset-generator/src/search_engine/interface.py
```diff
@@ -0,0 +1,30 @@
+from abc import ABC, abstractmethod
```
nseidan: by the way, just noticed, can we give more descriptive name for this file ? not just `interface.py ` which is very generic.

rre-dataset-generator/src/config.py
```diff
@@ -1,8 +1,16 @@
...
+
 from typing import List, Optional, Literal, Dict
 from pydantic import BaseModel, HttpUrl, Field, field_validator, FilePath
 import yaml
+import logging
 from pathlib import Path
 
+from src.logger import configure_logging
+
+configure_logging(level=logging.DEBUG)
```
nseidan, line 11: I'd prefer logging.INFO instead of DEBUG, @dantuzi @Intrinsical-AI WDYT ?

rre-dataset-generator/src/config.py
nicolo-rinaldi: fixed

rre-dataset-generator/src/search_engine/interface.py
nicolo-rinaldi: @nseidan `search_engine_interface.py` would be better?

nseidan: maybe` search_engine_base.py` as the class acts as a base for all supported engines?

nseidan (approved):

rre-dataset-generator/src/search_engine/interface.py
nicolo-rinaldi: changed

nseidan:
@nicolo-rinaldi while browsing files,  i see `try_solr.py` and` dataset_generator.py` files. So is dataset_generator main entrance to extract dataset and call LLM service? Also why we need a separate `try_solr.py`?    I am just questioning myself, because in the work item that i am doing I need a class where I can get solr data and call LLM service for queries extraction with that data.

nicolo-rinaldi:
The `try_solr.py` file is there just to test the connection to the Solr instance without the use of mocking (that has been used in integration tests). And also to highlight that also ticket [DAGE-7](https://sease.atlassian.net/browse/DAGE-7) has been done

rre-dataset-generator/try_solr.py
```diff
@@ -0,0 +1,27 @@
...
+    config = Config.load(args.config_file)
+
+    search_engine = SolrSearchEngine('http://localhost:8983/solr/testcore/')
+
+    # docs = search_engine.fetch_for_query_generation(documents_filter=config.documents_filter,
+    #                                                            doc_number=config.doc_number)
+
+    docs = search_engine.fetch_for_evaluation(keyword="and", query_template=config.query_template)
+
+    with open("solr_extracted_dataset.json", "w") as file:
```
nseidan: actually we dont need this file, instead we should put this code into dataset_generator.py, without writing into a file. Then LLMService would grab this docs  for query generation.

rre-dataset-generator/dataset_generator.py
```diff
@@ -1,26 +1,21 @@
...
-    configure_logging(level=logging.DEBUG)
-    log = logging.getLogger(__name__)
+    docs = search_engine.fetch_for_evaluation(keyword="and", query_template=config.query_template)
 
-    try:
-        config = load_config(args.config_file)
-        log.debug("Configuration loaded successfully.")
-    except Exception as e:
-        log.debug(f"Error loading configuration: {e}")
+    with open("solr_extracted_dataset.json", "w") as file:
```
nseidan: we don't need to store the extracted docs into a separate file. It is enough to check if docs  are not empty, have values; they are already stored in search engine.  It would be redundant.

rre-dataset-generator/solr_extracted_dataset.json
dantuzi, line 1: I don't think we need to commit this

rre-dataset-generator/src/search_engine/interface.py
nicolo-rinaldi: It throws a ValidationError if the class find a non-url like string. I added a test about that.

rre-dataset-generator/solr_extracted_dataset.json
nicolo-rinaldi: removed

rre-dataset-generator/try_solr.py
nicolo-rinaldi: Changed to be all inside `dataset_generator.py`

rre-dataset-generator/dataset_generator.py
nicolo-rinaldi: removed

rre-dataset-generator/src/config.py
Intrinsical-AI: I'd rather to keep all in DEBUG for now. I think it should be more atomic than INFO. 

DEBUG it's intented to support information for both: devs and clients. INFO, are intended to help only clients / users. So I'd keep debug till the end of the first sprint?

rre-dataset-generator/src/config.py
```diff
@@ -25,31 +33,36 @@ class Config(BaseModel):
     @field_validator('doc_fields')
     def check_no_empty_fields(cls, v):
         if any(not f.strip() for f in v):
+            log.error("docFields cannot contain empty strings.")
```
Intrinsical-AI, line 36: not mandatory, but maybe would be nice to store the strings into vars, and reuse them. Small savings on space and time!

rre-dataset-generator/src/config.py
```diff
@@ -25,31 +33,36 @@ class Config(BaseModel):
...
 
+    @classmethod
+    def load(cls, config_path: str) -> Config:
+        """
+        Load and validate configuration from a YAML file.
 
-def load_config(config_path: str) -> Config:
-    """
-    Load and validate configuration from a YAML file.
+        :param config_path: Path to the YAML config file
```
Intrinsical-AI, line 61: Only redability: I would keep consistency between the Docstrings. If we use :param, we should use in all the functions, and don't mix with other formats as:
"
        Args:
            template_payload (str): The Solr query string, e.g., 'q=ghosts&fq=genre:horror&wt=json'.
        Returns:
            dict: A dictionary representing the query parameters."
            
It's not critical ofc

rre-dataset-generator/src/config.py
```diff
@@ -25,31 +33,36 @@ class Config(BaseModel):
...
 
-    :param config_path: Path to the YAML config file
-    :return: Parsed and validated Config object
-    """
-    with open(config_path, 'r') as f:
-        raw_config = yaml.safe_load(f)
-    return Config(**raw_config)
\ No newline at end of file
+        log.debug("Configuration file loaded successfully.")
+        return cls(**raw_config)
```
Intrinsical-AI, line 68: Nice one!

rre-dataset-generator/src/search_engine/solr_search_engine.py
```diff
@@ -0,0 +1,120 @@
...
+            raise Timeout(f"Request to {search_url} timed out\nError: {e}")
+        except RequestException as e:
+            log.error(f"Unexpected error during request to {search_url}\nError: {e}")
+            raise RequestException(f"Unexpected error during request to {search_url}\nError: {e}")
+
+        match response.status_code:
+            case 200:
+                log.debug("Solr query successful.")
+                log.debug(f"URL: {search_url}\n")
+                log.debug(f"Payload: {payload}\n")
```
Intrinsical-AI, line 109: Redability: I'd make a single string in one line, with the 3 sentences, playing with \n if needed to better visualization

rre-dataset-generator/tests/unit/test_config.py
```diff
@@ -1,13 +1,13 @@
...
 from pathlib import Path
 from pydantic_core import ValidationError
+import pytest
 
+from src.config import Config
 
 @pytest.fixture
 def config():
-    return load_config("tests/unit/resources/good_config.yaml")
+    return Config.load("tests/unit/resources/good_config.yaml")
```
Intrinsical-AI, line 10: Nice :+1: x2

Intrinsical-AI (approved):
Redability improvements - neither of them critical.

dantuzi (approved):

PR #156, opened 2025-07-15
DAGE-9: Add in-memory data structure

nseidan:
* Added DataStore class to handle storing docs, queries, and scores. 
* Added model QueryRatingContext model 
* https://sease.atlassian.net/browse/DAGE-9


Intrinsical-AI (approved):

rre-dataset-generator/src/search_engine/data_store.py
```diff
@@ -0,0 +1,106 @@
+from __future__ import annotations
+
+import uuid
+from typing import Dict, Tuple, List
+
```
Intrinsical-AI, line 12: from src.logger import configure_logging
import logging

??

rre-dataset-generator/src/search_engine/data_store.py
```diff
@@ -0,0 +1,106 @@
...
+    Public API for storing, managing documents, queries, and relevance scores.
+    """
+
+    def __init__(self):
+        self.documents: Dict[str, str] = {}  # doc_id → document text
+        self.queries: Dict[str, QueryObject] = {}  # query_id → QueryObject
+
+    def _get_query_object(self, query_id: str) -> QueryObject:
+        if query_id not in self.queries:
+            raise KeyError(f"Query id '{query_id}' not found in DataStore.")
```
Intrinsical-AI: log.error / debug?

rre-dataset-generator/src/search_engine/data_store.py
```diff
@@ -0,0 +1,106 @@
...
+        return self.queries[query_id]
+
+    def _get_document(self, doc_id: str) -> str:
+        if doc_id not in self.documents:
+            raise KeyError(f"Document id '{doc_id}' not found in DataStore.")
```
Intrinsical-AI: log.error / debug?

rre-dataset-generator/src/search_engine/data_store.py
```diff
@@ -0,0 +1,106 @@
...
+    def add_score_for_query(self, doc_id: str, score: float) -> None:
+        self.doc_id_to_score[doc_id] = score
+
+    def has_score_for_query(self, doc_id: str) -> bool:
+        """
+        Returns True if this query has been scored for doc_id (i.e. score != -1) or raises KeyError
+        if the doc_id is not linked to this query.
+        """
+        if doc_id not in self.doc_id_to_score:
+            raise KeyError(f"Document id '{doc_id}' is not associated with this query.")
```
Intrinsical-AI: log.error / debug?

rre-dataset-generator/src/search_engine/data_store.py
Intrinsical-AI: I'd log every "raise" string context, into log.error / log.debug
E.g:
raise KeyError(f"Query id '{query_id}' not found in DataStore.")
We're missing relevant information of the failure in to our errors file!
I'd do like:
_error_str = f"Query id '{query_id}' not found in DataStore."
log.error(_error_str)
raise KeyError(_error_str)

Intrinsical-AI (commented):
Small improvements on logging

rre-dataset-generator/src/search_engine/data_store.py
nseidan: added logging

rre-dataset-generator/src/search_engine/data_store.py
nseidan: added

rre-dataset-generator/src/search_engine/data_store.py
nseidan: added

rre-dataset-generator/src/search_engine/data_store.py
nseidan: added

rre-dataset-generator/src/search_engine/data_store.py
```diff
@@ -0,0 +1,115 @@
...
+log = logging.getLogger(__name__)
+
+
+class DataStore:
+    """
+    Public API for storing, managing documents, queries, and relevance scores.
+    """
+
+    def __init__(self):
+        self.documents: Dict[str, str] = {}  # doc_id → document text
```
nicolo-rinaldi: In here only a string (document text) is saved. Why is that?

rre-dataset-generator/src/search_engine/data_store.py
```diff
@@ -0,0 +1,115 @@
...
+    def _get_document(self, doc_id: str) -> str:
+        if doc_id not in self.documents:
+            log.error("Document id %s not found in DataStore", doc_id)
+            raise KeyError(f"Document id '{doc_id}' not found in DataStore")
+        return self.documents[doc_id]
+
+    def add_document(self, doc_id: str, document: str) -> None:
+        self.documents[doc_id] = document
+
+    def get_document(self, doc_id: str) -> str:
```
nicolo-rinaldi: why don't we just change the protected version of this function? Not complaining, just curious about the motivation behind that

rre-dataset-generator/src/search_engine/data_store.py
```diff
@@ -0,0 +1,115 @@
...
+        """
+        return self._get_document(doc_id)
+
+    def add_query(self, query: str, doc_id: str) -> str:
+        """
+        Returns generated query id
+        """
+        query_obj = QueryObject(query, doc_id)
+        self.queries[query_obj.id] = query_obj
+        return query_obj.id
```
nicolo-rinaldi: why do we have the id back from this function?

rre-dataset-generator/src/search_engine/data_store.py
nseidan: we store "description" of the document here which is ultimatelty used in LLM to generate queries, @dantuzi correct me I could be wrong ?

rre-dataset-generator/src/search_engine/data_store.py
nseidan: protected version ?

rre-dataset-generator/src/search_engine/data_store.py
nseidan: for queries, it's better to manage with ids instead of plain query  text

rre-dataset-generator/src/search_engine/data_store.py
nicolo-rinaldi: the one with the _ at the beginning

nseidan: oh i see, it is just for internal details for validating first and re-used in few places.

rre-dataset-generator/src/search_engine/data_store.py
nicolo-rinaldi: yes, but why you decided to return a value instead of `None`?

nseidan: to store in queries dict
See dict with <query_id, QueryObject>

rre-dataset-generator/src/search_engine/data_store.py
```diff
@@ -0,0 +1,115 @@
...
+
+from src.logger import configure_logging
+
+configure_logging(level=logging.INFO)
+log = logging.getLogger(__name__)
+
+
+class DataStore:
+    """
+    Public API for storing, managing documents, queries, and relevance scores.
```
dantuzi: This is not an API

rre-dataset-generator/src/search_engine/data_store.py
dantuzi: In our configuration file, there was a parameter to specify the fields the script needs to use for generating the queries.
I'd probably store a more structured document. Maybe a dict[str, str] (field_name -> value) instead of a simple string.
I propose something like
```
self.documents: Dict[str, Dict[str, str]] = {}
```
What do you think?

rre-dataset-generator/src/search_engine/data_store.py
```diff
@@ -0,0 +1,115 @@
...
+    """
+
+    def __init__(self):
+        self.documents: Dict[str, str] = {}  # doc_id → document text
+        self.queries: Dict[str, QueryObject] = {}  # query_id → QueryObject
```
dantuzi: If the query is generated from multiple documents?
I'd say this should be 
```
self.queries: Dict[str, List[QueryObject]] = {}
```

rre-dataset-generator/src/search_engine/data_store.py
```diff
@@ -0,0 +1,115 @@
...
+
+    def _get_query_object(self, query_id: str) -> QueryObject:
+        if query_id not in self.queries:
+            log.error("Query id %s not found in DataStore", query_id)
+            raise KeyError(f"Query id '{query_id}' not found in DataStore")
+        return self.queries[query_id]
+
+    def _get_document(self, doc_id: str) -> str:
+        if doc_id not in self.documents:
+            log.error("Document id %s not found in DataStore", doc_id)
```
dantuzi: Why do you think this is an error? I don't have a strong opinion on this, but I would have logged a warning.

rre-dataset-generator/src/search_engine/data_store.py
```diff
@@ -0,0 +1,115 @@
...
+            raise KeyError(f"Document id '{doc_id}' not found in DataStore")
+        return self.documents[doc_id]
+
+    def add_document(self, doc_id: str, document: str) -> None:
+        self.documents[doc_id] = document
```
dantuzi: It doesn't check if the document ID already exists.
I would raise an exception if the document already exists. If we need to update a document in the future, we can create another method to update it explicitly

rre-dataset-generator/src/search_engine/data_store.py
```diff
@@ -0,0 +1,115 @@
...
+class QueryObject:
+    """
+    QueryObject: holds the query text, its doc id → score mapping, and a generated unique id.
+    """
+
+    def __init__(self, query: str, doc_id: str):
+        self.id: str = str(uuid.uuid4())
+        self.query: str = query
+        # mapping doc_id → score; score=-1 means doc is “not scored yet”
+        self.doc_id_to_score: Dict[str, float] = {doc_id: -1.0}
```
dantuzi: Instead of adding this line of documentation, I would rather create a constant NOT_SCORED_YET = -1

rre-dataset-generator/src/search_engine/data_store.py
```diff
@@ -0,0 +1,115 @@
...
+        Returns a document text or raises KeyError if the doc_id is not found.
+        """
+        return self._get_document(doc_id)
+
+    def add_query(self, query: str, doc_id: str) -> str:
+        """
+        Returns generated query id
+        """
+        query_obj = QueryObject(query, doc_id)
+        self.queries[query_obj.id] = query_obj
```
dantuzi: If the query already exists (for this first version, we compare if the query strings are the same), we shouldn't create a new entry, but we need to update the old one, adding a new docID

rre-dataset-generator/src/search_engine/data_store.py
dantuzi: After a Zoom chat and the [RR](https://github.com/SeaseLtd/rated-ranking-evaluator/pull/159), we decided to update this line to be

```
self.documents: Dict[str, Document] = {}
```

rre-dataset-generator/src/search_engine/data_store.py
```diff
@@ -0,0 +1,115 @@
...
+
+class QueryObject:
+    """
+    QueryObject: holds the query text, its doc id → score mapping, and a generated unique id.
+    """
+
+    def __init__(self, query: str, doc_id: str):
+        self.id: str = str(uuid.uuid4())
+        self.query: str = query
+        # mapping doc_id → score; score=-1 means doc is “not scored yet”
```
dantuzi: this is not a score but a rating which is an integer

rre-dataset-generator/src/search_engine/data_store.py
nseidan: updated

rre-dataset-generator/src/search_engine/data_store.py
nseidan: updated

rre-dataset-generator/src/search_engine/data_store.py
nseidan: updated

rre-dataset-generator/src/search_engine/data_store.py
nseidan: updated

rre-dataset-generator/src/search_engine/data_store.py
nseidan: updated

rre-dataset-generator/src/search_engine/data_store.py
nseidan: updated

rre-dataset-generator/src/search_engine/data_store.py
nseidan: updated

rre-dataset-generator/src/model/query_rating_context.py
```diff
@@ -0,0 +1,44 @@
...
+    """
+
+    DOC_NOT_RATED: int = -1  # doc is not yet rated
+
+    def __init__(self, query: str, doc_id: str):
+        self._id: str = str(uuid.uuid4())
+        self._query: str = query
+        self._doc_id_to_rating_score: Dict[str, int] = {doc_id: self.DOC_NOT_RATED}
+
+    def get_id(self) -> str:
```
dantuzi: This can be renamed to `get_query_id` to avoid confusion with the doc_id

rre-dataset-generator/src/search_engine/data_store.py
```diff
@@ -0,0 +1,103 @@
...
+        """
+        if query in self._query_text_to_query_id:
+            query_id = self._query_text_to_query_id[query]
+            context = self._queries_by_id[query_id]
+            context.add_doc_id(doc_id)
+            return query_id
+
+        # new query rating context
+        context = QueryRatingContext(query, doc_id)
+        context_id = context.get_id()
```
dantuzi: Why is this named `context_id` while above, the same value, is called `query_id`?
I'd call everything query_id

rre-dataset-generator/src/search_engine/data_store.py
```diff
@@ -0,0 +1,103 @@
...
+            raise KeyError(f"Document id '{doc_id}' found in DataStore")
+        self._documents[doc_id] = document
+
+    def get_document(self, doc_id: str) -> Optional[Document]:
+        """
+        Returns Document or None.
+        """
+        return self._get_document(doc_id)
+
+    def add_query(self, query: str, doc_id: str) -> str:
```
dantuzi, line 48: Can you write some unit test for this?

rre-dataset-generator/src/model/query_rating_context.py
nseidan: updated

rre-dataset-generator/src/search_engine/data_store.py
nseidan: updated

rre-dataset-generator/src/search_engine/data_store.py
nseidan: added

Intrinsical-AI (approved):
Let's unblock!

dantuzi (approved):
Just a minor to rename the test method names to follow the pattern
<what I'm testing>_expect_<what to expect>

nseidan:
> Just a minor to rename the test method names to follow the pattern <what I'm testing>_expect_

renamed

PR #158, opened 2025-07-15
DAGE-13: Add initial RRE investigation and setup

Intrinsical-AI:
I think DOGE-13.md it's worth to keep.

I added some simple docs to test.

Target folder (evaluation results), it's ommited by .gitignore

Works for me in local (for the mvn clean install part, I had to change some references on deprecated libraries, on the root pom.xml)


Intrinsical-AI:
This was moved to an expanded research tasks. Closing the PR.

PR #164, opened 2025-07-17
Adding --verbose param to CLI

nicolo-rinaldi:
Centralized all the `configure_logging` calls to be in the main file. This is because we want to have the same log level in each file.

### DIFF
--- a/rre-dataset-generator/src/utils.py
+++ b/rre-dataset-generator/src/utils.py
@@ -1,6 +1,8 @@
+import argparse
 import re
 import html
-import argparse
+
+_TAG_REGEX = re.compile('<.*?>')
 
 def parse_args():
     parser = argparse.ArgumentParser(description='Parse arguments for CLI.')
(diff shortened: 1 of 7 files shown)


dantuzi (dismissed):
I like it

rre-dataset-generator/src/utils.py
```diff
@@ -9,9 +9,12 @@ def parse_args():
                         help='Config file path to use for the application [default: \"config.yaml\"]',
                         required=False, default="config.yaml")
 
+    parser.add_argument('-v', '--verbose',action='store_true',
+                        help='Activate debug mode for logging [default: False]')
+
     return parser.parse_args()
 
 def clean_text(text):
     tag_regex = re.compile('<.*?>')
```
nseidan: currently this regex is compiled per call which is expensive, better to have it at the model/class level, so that it's compiled once:
`_TAG_REGEX = re.compile()`

rre-dataset-generator/src/utils.py
```diff
@@ -9,9 +9,12 @@ def parse_args():
...
     text_without_html = re.sub(tag_regex, '', text)
-    return html.unescape(re.sub("\s\s+", " ", text_without_html))
+    return html.unescape(re.sub(r"\s{2,}", " ", text_without_html))
```
nseidan, line 21: what about if we have trailing/leading spaces ?  -> `strip() `?

Intrinsical-AI (dismissed):

rre-dataset-generator/src/utils.py
nicolo-rinaldi: done

rre-dataset-generator/src/utils.py
nicolo-rinaldi: modified as discussed

nseidan (dismissed):

Intrinsical-AI (approved):

nseidan (approved):

dantuzi (approved):

PR #166, opened 2025-07-17
[DAGE-12] Added code to generate the scores (looking at Naz code)

nicolo-rinaldi:
## Changes
- Added function generate_scores in 'llm_service.py' to generate the score given a document a query text.
- Added tests at 'test_llm_service.py'
- typo in regex of clean_text function in 'utils.py'

[ticket](https://sease.atlassian.net/browse/DAGE-12)


rre-dataset-generator/src/llm/llm_service.py
```diff
@@ -21,3 +28,48 @@ def test_connection(self, prompt: str) -> str:
...
+            error_msg = "The relevance scale must be either 'binary' or 'graded'"
+            log.error(error_msg)
+            raise ValueError(error_msg)
+
+        messages = [
+            SystemMessage(
+                content=f"You are a professional data labeler and, given a documents with a set of fields and a query "
+                        f"text, you need to return the relevance score in a scale called {relevance_scale.upper()}. The "
+                        f"scores of this scale are built as follows:\n{description}\n"
+                        f"Knowing this, return a JSON object with key 'score' and the related score as an integer value."
```
Intrinsical-AI, line 68: I'd like to ask it explicitly to return a JSON with a key "score". 

I'd also add a mocked example into the prompt like

EXPECTED RESPONSE:
{"score":1}

Or smth similar

Intrinsical-AI (dismissed):

rre-dataset-generator/src/llm/llm_service.py
```diff
@@ -21,3 +28,49 @@ def test_connection(self, prompt: str) -> str:
 
         response = self.chat_model.invoke(messages)
         return response.content.strip()
+
+    def generate_scores(self, document: Document, query: str, relevance_scale: str) -> int:
+        """
+        Send a test prompt to verify the model returns a response.
```
nseidan: not a test prompt ?

rre-dataset-generator/src/llm/llm_service.py
```diff
@@ -21,3 +28,49 @@ def test_connection(self, prompt: str) -> str:
...
+        """
+        if relevance_scale == "binary":
+            scale = {0, 1}
+            description = (" - 0: the query is NOT relevant looking at the document"
```
nseidan: I would simplify the description: `0: the query is NOT relevant to the given document`

nseidan:
@nicolo-rinaldi  have you manually tested  this with openai ? We have the openai api key . Asking because your system message starts with `You are a professional data labeler and, given a documents with`, and wondering what response would get from the model?

rre-dataset-generator/src/llm/llm_service.py
nicolo-rinaldi: ops, typo. Changed

nicolo-rinaldi:
Yes, I tried different times with some cases. Seems to work as expected

rre-dataset-generator/src/llm/llm_service.py
nicolo-rinaldi: done

nseidan (dismissed):

rre-dataset-generator/src/llm/llm_service.py
nicolo-rinaldi: Is it ok if we start just by giving an example of the output format? Then, if we found that is working better with few-shot prompting, we can easily change that.

Intrinsical-AI (dismissed):

rre-dataset-generator/src/llm/llm_service.py
```diff
@@ -42,3 +41,49 @@ def generate_queries(self, document: Document, num_queries_generate_per_doc: int
...
+            log.error(error_msg)
+            raise ValueError(error_msg)
+
+        messages = [
+            SystemMessage(
+                content=f"You are a professional data labeler and, given a documents with a set of fields and a query "
+                        f"text, you need to return the relevance score in a scale called {relevance_scale.upper()}. The "
+                        f"scores of this scale are built as follows:\n{description}\n"
+                        f"Knowing this, return a JSON object with key 'score' and the related score as an integer value."
+                        f"I'm expecting a response like the following: {{\"score\": `integer`}}"
```
dantuzi: How about if we write "I'm expecting a JSON response like... "
In this case, we specify the output format, and we ask for a JSON-compliant response. What do you think?

rre-dataset-generator/src/llm/llm_service.py
```diff
@@ -42,3 +41,49 @@ def generate_queries(self, document: Document, num_queries_generate_per_doc: int
         except JSONDecodeError:
             log.warning("LLM hallucinated and its response: %s", raw)
             return raw
+
+    def generate_scores(self, document: Document, query: str, relevance_scale: str) -> int:
```
dantuzi: I would get a type ENUM for the relevance_scale instead of a string parsed and validated directly when we read the config file

rre-dataset-generator/src/llm/llm_service.py
```diff
@@ -42,3 +41,49 @@ def generate_queries(self, document: Document, num_queries_generate_per_doc: int
...
+            response =  int(json.loads(raw)['score'])
+            if response not in scale:
+                error_msg = f"LLM hallucinated the value of the scale. Returned: {response}"
+                log.warning(error_msg)
+                raise ValueError(error_msg)
+            return response
+        except JSONDecodeError:
+            error_msg = f"LLM hallucinated and its response: {raw}"
+            log.warning(error_msg)
+            raise ValueError(error_msg)
```
dantuzi: new line

rre-dataset-generator/src/llm/llm_service.py
```diff
@@ -42,3 +41,49 @@ def generate_queries(self, document: Document, num_queries_generate_per_doc: int
...
+                        f"Knowing this, return a JSON object with key 'score' and the related score as an integer value."
+                        f"I'm expecting a response like the following: {{\"score\": `integer`}}"
+            ),
+            HumanMessage(
+                content=f"Document: {document.model_dump_json()}\n"
+                        f"Query:{query}\n"
+            )
+        ]
+
+        #response = self.chat_model.with_structured_output(method="json_mode").invoke(messages)
```
dantuzi, line 77: do we need this commented line?

rre-dataset-generator/src/utils.py
```diff
@@ -16,6 +16,6 @@ def parse_args():
 
     return parser.parse_args()
 
-def clean_text(text):
+def clean_text(text: str) -> str:
```
dantuzi, line 19: good catch

rre-dataset-generator/src/llm/llm_service.py
dantuzi: it should be `generate_score` as it generates a single score

rre-dataset-generator/tests/unit/resources/solr_config.yaml
dantuzi, line 1: Why do we need this?
Is it possible this is a leftover of the task you are currently working on?

rre-dataset-generator/src/llm/llm_service.py
nicolo-rinaldi: It's safer way to have a structured output. Is not supported for each model. I kept this line to remind myself that it is another way of doing the same thing. Maybe, with some LangChain update all the models that **dataset-generator** supports might have this functionality, and we can change this line of code

rre-dataset-generator/src/llm/llm_service.py
nicolo-rinaldi: true, solved

rre-dataset-generator/tests/unit/resources/solr_config.yaml
nicolo-rinaldi: Yes, it was

rre-dataset-generator/src/llm/llm_service.py
nicolo-rinaldi: yeah, it makes sense to specify again. Changed

rre-dataset-generator/src/llm/llm_service.py
nicolo-rinaldi: Do you want me to change the structure of the `Config` class to store the Enum version of the relevance scale?

dantuzi: I think it's cleaner but we can discuss offline and possibly modifying it in a separate pr

dantuzi (approved):

PR #171, opened 2025-07-18
Dage 16: response validation Schemas

Intrinsical-AI:
## TASK TICKET
[DAGE-16](https://sease.atlassian.net/jira/software/projects/DAGE/boards/84?selectedIssue=DAGE-16) – Introduces Pydantic model LLMScoreResponse and refactors LLMService to enforce structured output and "free" validation / checks.

### Last update: back to basics

We decide to use simple data-structures (dicts / lists)  for simplicity / efficiency. Pydantic it's (yet) an over kill - but might come back around.

-> Back to simple approaches. Used simple data structures to perform validation of LLM responses in the LLM Service

----
#### Changes

- Added pydantic schemas for LLM service. Needed to re-write the generate_queries output.
- Added LLMScoreResponse schema with score validation logic
- Refactored generate_score() to use structured parsing and stricter output checks
- Split and extended tests into tests/unit/llm/ for generate_score() and generate_queries()
- Parametrized valid/invalid score scenarios, malformed JSON, and hallucinated outputs
- Improved validation error handling and logging


rre-dataset-generator/src/model/llm_schemas.py
```diff
@@ -0,0 +1,76 @@
...
+            raise ValueError("All items in `content` must be strings")
+
+        if any(item.strip() == "" for item in parsed):
+            raise ValueError("Empty or whitespace-only strings are not allowed in `content`")
+
+        self.content_list = parsed
+        return self
+
+
+## NOT NEEDED YET - WOULD BE USEFUL TO ADD PYDANTIC VALIDATION TO THE LLM Service
```
nseidan: if not needed now, then i'd  remove all these comments; we could introduce when needed.

rre-dataset-generator/src/model/llm_schemas.py
```diff
@@ -0,0 +1,76 @@
...
+
+import json
+from typing import List
+from pydantic import BaseModel, Field, model_validator
+
+# Typing the LLM query service output 
+class LLMQueryResponse(BaseModel):
+    """Output model for LLM service responses."""
+    content: str = Field(..., description="The generated response content as JSON string")
+    content_list: List[str] = Field(default_factory=list, description="Parsed content as a list of strings")
```
nseidan: why need to store both: `content` & `content_list`?  -> redundancy

rre-dataset-generator/src/llm/llm_service.py
```diff
@@ -34,11 +36,22 @@ def generate_queries(self, document: Document, num_queries_generate_per_doc: int
...
-            return json.loads(raw)
-        except JSONDecodeError:
-            log.warning("LLM hallucinated and its response: %s", raw)
-            return raw
+        # The response from invoke is an AIMessage object which contains all the needed info
+        response = self.chat_model.invoke(messages)
+
+        # Extract token usage and model name from response metadata
+        # usage = response.response_metadata.get("token_usage", {})
+        # model_name = response.response_metadata.get("model_name", "unknown")
```
nseidan: why need `model_name` ? It is passed by the user ?

rre-dataset-generator/src/llm/llm_service.py
```diff
@@ -34,11 +36,22 @@ def generate_queries(self, document: Document, num_queries_generate_per_doc: int
...
+        # finish_reason = response.response_metadata.get("finish_reason", "unknown")
+
+        output = LLMQueryResponse(
+            content=response.content,
+            # model=model_name,
+            # usage={
+            #     "prompt_tokens": usage.get("prompt_tokens", 0),
+            #     "completion_tokens": usage.get("completion_tokens", 0),
+            #     "total_tokens": usage.get("total_tokens", 0),
```
nseidan: why all those are commented out ?

rre-dataset-generator/src/model/llm_schemas.py
Intrinsical-AI: I rather keep it as a comment that we can always delete, than having to re-implement all the logic afterwards? Or how would you do it?

rre-dataset-generator/src/model/llm_schemas.py
Intrinsical-AI: Was requested by @dantuzi content it's the raw LLM output, while content_list are each of the individual queries, in a list format. Both are check / validated with Pydantic in instantiation time

rre-dataset-generator/src/llm/llm_service.py
Intrinsical-AI: The idea it's keeping all the API "calling" metadata cerntalized in the response object

rre-dataset-generator/src/llm/llm_service.py
Intrinsical-AI: Because are not needed right away, but could be useful in the future -> 
"The idea it's keeping all the API "calling" metadata cerntalized in the response object"
We can calculate inference price afterwards, or even add attributes as "latency" / "generation time" / timestamps

rre-dataset-generator/src/model/llm_schemas.py
nicolo-rinaldi: I don't know if we will ever add this feature, since we are creating ourself the message to send. Anyway, I would rather remove this comment and keep them in a local folder as a file out of the repo, to avoid too much things inside the file. This might be misleading on who reads this file

rre-dataset-generator/src/llm/llm_service.py
nicolo-rinaldi: I'd rather remove the comments

rre-dataset-generator/src/llm/llm_service.py
```diff
@@ -34,11 +36,22 @@ def generate_queries(self, document: Document, num_queries_generate_per_doc: int
             HumanMessage(content=f"Document:\n{doc_json}")
         ]
 
-        raw = self.chat_model.invoke(messages).content.strip()
-
-        # in case LLM hallucinates
-        try:
-            return json.loads(raw)
-        except JSONDecodeError:
-            log.warning("LLM hallucinated and its response: %s", raw)
```
nicolo-rinaldi, line 44: Log is missing in the new file

nicolo-rinaldi (commented):
I left just minor comments. I'd add some logging along the way. The comments are a bit ugly to look at. Maybe is better to keep them in separate files(?)

rre-dataset-generator/src/llm/llm_service.py
nseidan: I'd remove those. See https://en.wikipedia.org/wiki/You_aren%27t_gonna_need_it

nseidan (dismissed):

rre-dataset-generator/src/llm/llm_service.py
dantuzi: the log is now in the new version. Resolving

rre-dataset-generator/src/llm/llm_service.py
```diff
@@ -34,57 +39,56 @@ def generate_queries(self, document: Document, num_queries_generate_per_doc: int
...
+        except (KeyError, JSONDecodeError, ValueError) as e:
+            log.warning(f"LLM unexpected response. Raw output: {raw}")
+            raise ValueError(f"Invalid LLM response: {e}")
+
+        if parsed.score not in allowed:
+            msg = f"Score '{parsed.score}' is not valid in {relevance_scale} scale"
+            log.warning(msg)
+            raise ValueError(msg)
+
+        return parsed.score
```
dantuzi: new line

rre-dataset-generator/src/model/llm_schemas.py
```diff
@@ -0,0 +1,63 @@
...
+
+import json
+from typing import List
+from pydantic import BaseModel, Field, model_validator, field_validator
+
+# ───────────────────────────────────────────────────────────────
+# Response model for LLM-generated output
+# ───────────────────────────────────────────────────────────────
+
+class LLMQueryResponse(BaseModel):
```
dantuzi: not blocking, but I don't see why you like to put all classes in the same file

rre-dataset-generator/src/model/llm_schemas.py
```diff
@@ -0,0 +1,63 @@
...
+        score (int): Relevance score returned by the model.
+    """
+    score: int = Field(..., description="Relevance score returned by LLM.")
+
+    @field_validator("score")
+    @classmethod
+    def validate_score(cls, value):
+        if value not in {0, 1, 2}:
+            raise ValueError("Score must be 0, 1, or 2.")
+        return value
```
dantuzi: new line

rre-dataset-generator/src/model/llm_schemas.py
```diff
@@ -0,0 +1,63 @@
...
+    
+    Attributes:
+        content (str): Raw content generated by the LLM as a JSON-encoded string.
+        content_list (List[str]): Parsed list of strings extracted from `content`. -> This is the "practical" usage field
+    """
+    content: str = Field(..., description="Raw response content as a JSON string representing a list of strings.")
+    content_list: List[str] = Field(default_factory=list, description="Parsed list of strings from the JSON content.")
+
+    @model_validator(mode="after")
+    def validate_and_parse_content(self) -> 'LLMQueryResponse':
```
dantuzi: Is this triggered every time the object is updated? How is the object initialized?
We should normally start from our needs and choose the technologies, not choose a technology and adapt it to meet our needs.
Reading this class, I have the impression that we are adapting the Pydantic feature to parse and validate a JSON.

Look at the following, it is just a copy/paste of your code, but it implements a simple class with a private and non-modifiable attribute
```
class LLMQueryResponse:
    _queries: List[str]

   def __init__(response:str):
        try:
            self._queries = json.loads(response)
        except json.JSONDecodeError as e:
            raise ValueError(f"Invalid JSON in `content`: {e}")

        if not isinstance(self._queries, list):
            raise ValueError("`content` must be a JSON list.")

        if not all(isinstance(item, str) for item in self._queries):
            raise ValueError("All items in `content` must be strings.")

        if any(not item.strip() for item in parsed):
            raise ValueError("Items in `content` must not be empty or only whitespace.")

    def get_queries()
         return self._queries
       
```

rre-dataset-generator/src/model/llm_schemas.py
```diff
@@ -0,0 +1,63 @@
...
+    
+
+class LLMScoreResponse(BaseModel):
+    """
+    Output model for responses from the LLM scoring function.
+
+    Attributes:
+        score (int): Relevance score returned by the model.
+    """
+    score: int = Field(..., description="Relevance score returned by LLM.")
```
dantuzi: Same as before, it can be a simple class with a single non-modifiable attribute

rre-dataset-generator/src/model/llm_schemas.py
Intrinsical-AI: 1) This only executes after intanciation of the object (luckily). - ( I think the difference with @field_validator might be that it's executed once all fields are validated and the object has been partially created. So you can make cross-comparisons, or side effects as this case)
2) LLMQueryResponse(llm_response_str)

In this case, I 100% agree with you. For the actual implementation, Pydantic it's an overkill, and (adds some extra costs on exec). So I'd rather preffer your approach currently. 

That being said there's some things why I thinked it could (assumed would) be useful:

- For the metadata and traceability 
- Which model generated that reponse? 
- When was it (the same model get's adjusted with time)?
- Cost checking -> token_price dictionay + output_tokens = cost of the response
- For logging -> As Pydantic it's default serializable, it's possible to get descriptive logging easily.

But you're righ, rn this is an overkill, probably.

Thanks for the feedback!

rre-dataset-generator/src/model/llm_schemas.py
Intrinsical-AI: You'remaking me curious about some experiment comparison on the increasing of resource consumption given by pydantic . What's the real overhead? And it's worth vs the extensibility potential?

rre-dataset-generator/tests/unit/llm/test_llm_service_queries.py
```diff
@@ -0,0 +1,54 @@
...
+    return Document(
+        id="doc1",
+        fields={
+            "title": "Car of the Year",
+            "description": "The Toyota Camry, the nation's most popular car has now been rated as its best new model."
+        }
+    )
+
+
+def test_llm_service_generate_queries_valid(example_doc):
```
dantuzi: ... _EXPECT_....

rre-dataset-generator/tests/unit/llm/test_llm_service_score.py
```diff
@@ -0,0 +1,53 @@
...
+    return Document(
+        id="doc1",
+        fields={
+            "title": "Car of the Year",
+            "description": "The Toyota Camry, the nation's most popular car has now been rated as its best new model."
+        }
+    )
+
+
+@pytest.mark.parametrize("scale, valid_score", [
```
dantuzi, line 21: Nice

rre-dataset-generator/src/model/query_response.py
```diff
@@ -0,0 +1,40 @@
+import json
+from typing import List
+
+class LLMQueryResponse:
+    """
+    Parses and validates a JSON response containing a list of queries.
+    """
+    _queries: List[str]
```
nseidan, line 8: @Intrinsical-AI while looking into LLM explainability feature work, we should introduce a new class, for example QueryExplanation and pass it here instead of just string:
```
_queries: List[QueryExplanation]
```

```
class QueryExplanation:
    query: str
    score: Optional[int] = None
    reasoning: Optional[str] = None
```
But if you feel it is out of this task scope, it's fine. I will introduce later. Just mentioning here.

Intrinsical-AI: Hello Naz! You're right: QueryExplaination might be also needed in the future!

In general would be much simplier if we store information as objects with simpler fields, but by now I personally would keep it in a datastructure to align the rest of the class. As it's your task, for my part feel free to choose.

 Check the proposal to disk persistance, to get a deeper insight: think reasoning could be implementable with a similar pattern / logic:
https://github.com/SeaseLtd/rated-ranking-evaluator/pull/179

:smile:

dantuzi: @nseidan please introduced the new class as part of the explainability PR so everything is self contained

dantuzi (approved):

rre-dataset-generator/src/llm/llm_service.py
```diff
@@ -35,32 +35,35 @@ def generate_queries(self, document: Document, num_queries_generate_per_doc: int
...
-                           " - 1: the query is relevant to the given document")
+            allowed = {0, 1}
+            description = (" - 0: the query is NOT relevant to the given document\n"
+                        " - 1: the query is relevant to the given document")
         elif relevance_scale == "graded":
-            scale = {0, 1, 2}
-            description = (" - 0: the query is NOT relevant to the given document"
-                           " - 1: the query may be relevant to the given document"
-                           " - 2: the document proposed is the answer to the query")
+            allowed = {0, 1, 2}
```
nseidan, line 59: `allowed` is assigned but never used.

rre-dataset-generator/src/model/score_response.py
```diff
@@ -0,0 +1,33 @@
...
+            raise ValueError(f"Score must be 0, 1, or 2 for graded scale, got {score}")
+            
+        self.score = score
+
+    def get_score(self) -> int:
+        """
+        Returns the validated score.
+
+        Returns:
+            The validated score.
```
nseidan, line 31: duplicate description

rre-dataset-generator/src/llm/llm_service.py
Intrinsical-AI: Hello Naz, you're right. Feel free to update as discussed!

rre-dataset-generator/src/model/score_response.py
Intrinsical-AI: Will bring up the Docstring unification in next standups

PR #172, opened 2025-07-21
DAGE-14: Integration tests - working on Docker + Solr testing

Intrinsical-AI:
- Relocated some files + cleaning (Docker Compose to project root folder; solr-init also)

- Test stack added

- Added tests/integration/docker-compose.yml (Solr 9.8.1 + solr-precreate testcore) to keep CI fast and isolated. Simpler than the original

- Pytest fixtures (tests/conftest.py):
+ docker_compose_file – tells pytest-docker where the compose lives.
+ solr_url – waits for testcore to be responsive and returns the base URL.
+ seed_dataset – idempotently seeds dataset.json when the core is empty.

- Integration tests (tests/integration/test_docker_solr.py)

See: https://sease.atlassian.net/browse/DAGE-14?focusedCommentId=16343


rre-dataset-generator/tests/integration/test_docker_solr.py
```diff
@@ -0,0 +1,25 @@
+import pytest
+import requests
+from src.search_engine.solr_search_engine import SolrSearchEngine
+
+def test_core_exists(solr_url):
+    r = requests.get(solr_url + "../admin/cores?action=STATUS&wt=json")
```
nicolo-rinaldi, line 6: what is this `../`?

rre-dataset-generator/tests/integration/docker-compose.yml
nicolo-rinaldi, line 1: Why do we need to change this? what are the benefits?

rre-dataset-generator/tests/conftest.py
```diff
@@ -0,0 +1,54 @@
...
+    url  = f"http://{docker_ip}:{port}/solr/testcore"
+
+    def _is_ready() -> bool:
+        try:
+            r = requests.get(url + "/admin/ping?wt=json")
+            return r.status_code == 200
+        except ConnectionError:
+            return False
+
+    docker_services.wait_until_responsive(
```
nicolo-rinaldi, line 29: This solr_url function is called each time the variable with the same name is passed to each test function?

rre-dataset-generator/tests/conftest.py
```diff
@@ -0,0 +1,54 @@
...
+        timeout=60, pause=0.5, check=_is_ready
+    )
+    return url + "/"   
+
+
+
+
+@pytest.fixture(scope="session", autouse=True)
+def seed_dataset(solr_url):
```
nicolo-rinaldi, line 38: Why are we updating the solr instance if we don't find the docs? shouldn't we only throw an Error?

rre-dataset-generator/tests/integration/resources/llm_config.yaml
```diff
@@ -0,0 +1,3 @@
+  # tests/integration/resources/llm_config.yaml
+  model: dummy
+  api_key: dummy
```
nicolo-rinaldi: missing whiteline

rre-dataset-generator/tests/integration/docker-compose.yml
```diff
@@ -1,15 +1,10 @@
...
-    build:
-      context: solr-init
-    depends_on:
-      - solr
-    command: ["sh", "-c", "/opt/rre-dataset-generator/solr-init.sh"]
-
+    healthcheck:
+      test: ["CMD", "curl", "-f", "http://localhost:8983/solr/testcore/admin/ping"]
+      interval: 5s
+      retries: 10
```
nicolo-rinaldi: missing empty line

rre-dataset-generator/tests/integration/test_docker_solr.py
```diff
@@ -0,0 +1,25 @@
...
+@pytest.mark.parametrize("field", ["id", "title"])
+def test_docs_have_field(solr_url, field):
+    r = requests.get(solr_url + "select?q=*:*&rows=1&fl=id,title&wt=json")
+    assert r.ok
+    assert field in r.json()["response"]["docs"][0]
+
+
+def test_search_engine_fetch(solr_url):
+    engine = SolrSearchEngine(solr_url)
+    docs   = engine.fetch_for_query_generation(None, 3, ["title","body"])
```
nicolo-rinaldi, line 24: I think we don't have "body" in our solr documents

rre-dataset-generator/tests/conftest.py
Intrinsical-AI: -> From what I understand no, but instead 1 time at the beggining of the execution with pytests tests/.. when the "conftest.py" get's "loaded"

rre-dataset-generator/tests/conftest.py
Intrinsical-AI: No, for the tests I think we would also want to also check document indexing. 

This fixture, loads the testing dataset configured by @nseidan into the Solr instance, only one time, and then you clone the instance (only using the variable seed_dataset), and "inyect it" into each test. Did I understand your question correctly?

rre-dataset-generator/tests/integration/docker-compose.yml
Intrinsical-AI: The last version was a bit  heavy for the pytest (as we decided to do the migration to a specilized python library), I assume it's supposed to be the real docker-compose.yml for the application, in that case that version should live in the root folder, and we need to add a simpler version for the testing in the integration folder. That's what I've done. I'm sure we can discuss a better solution. Check:
https://sease.atlassian.net/jira/software/projects/DAGE/boards/84?selectedIssue=DAGE-14
and feel free to ping me on Slack!

rre-dataset-generator/tests/integration/test_docker_solr.py
Intrinsical-AI: Right

Intrinsical-AI: I did not realize due to silent behaviour coming from engine.fetch_for_query_generation - that accept every string as field, even it's not registered in the index, and does not lof any warning message / error nor exception

rre-dataset-generator/tests/integration/test_docker_solr.py
Intrinsical-AI: solr_url  # → "http://127.0.0.1:8983/solr/testcore/"


But we want to access: "http://127.0.0.1:8983/solr/admin/cores?action=STATUS"

Web directory traversal: .. it's the same as cd .. in Bash

 We're going up 1 level in the path

rre-dataset-generator/tests/integration/resources/llm_config.yaml
Intrinsical-AI: solved

rre-dataset-generator/tests/integration/docker-compose.yml
Intrinsical-AI: solved

rre-dataset-generator/docker-compose.yml
```diff
@@ -0,0 +1,14 @@
...
+    ports:
+      - "8983:8983"
+    command: ["solr-precreate", "testcore"]
+
+  solr-init:
+    build:
+      context: solr-init
+    depends_on:
+      - solr
+    command: ["sh", "-c", "/rre-dataset-generator/solr-init/solr-init.sh"]
```
nseidan, line 14: @Intrinsical-AI we shouldn't move the docker-compose file to a root folder, it's a test env; not a service we produce, it is coming from user side; also our project is not a docker-based service, it is just CLI, no need to put it into the root folder.  cc @dantuzi

rre-dataset-generator/tests/integration/docker-compose.yml
```diff
@@ -1,15 +1,11 @@
...
     command: ["solr-precreate", "testcore"]
-
-  solr-init:
-    build:
-      context: solr-init
-    depends_on:
-      - solr
-    command: ["sh", "-c", "/opt/rre-dataset-generator/solr-init.sh"]
+    healthcheck:
+      test: ["CMD", "curl", "-f", "http://localhost:8983/solr/testcore/admin/ping"]
```
nseidan, line 8: @Intrinsical-AI  we don't need a healthcheck here, already handled in `solr-init.sh`

rre-dataset-generator/docker-compose.yml
Intrinsical-AI: Hey Naz, hanks for the feedback! this is a research task. It's a proposal for using python-docker.

 I had already tagged you in the Jira explaining it.
https://sease.atlassian.net/browse/DAGE-14?focusedCommentId=16413

 Please, check the board before review! :+1:

rre-dataset-generator/tests/integration/docker-compose.yml
Intrinsical-AI: 1) This research proposal suggest using python-docker, the one you mentioned during a call. if so, the checks of solr-init.sh will be moved to a proper pytests file that uses fixtures. The information it's here:  https://sease.atlassian.net/browse/DAGE-14?focusedCommentId=16413
2) This is not yet accepted - it's the proposal on "how" to do Integration tests of Solr - Docker.
3) In general docker compose would be better to be on the root.
4) in general: it's a good practice to implement healthheck on the docker compose, even you "handle" it outside (too).

I guess we will have to discuss in detail, but if you can propose a solution to easy orchestrating all the tests, would be nice!

rre-dataset-generator/tests/integration/test_docker_solr.py
nicolo-rinaldi: why we want to access that url? I was assuming that we test if the specific `testcore` was in the collection

rre-dataset-generator/tests/integration/test_docker_solr.py
nicolo-rinaldi: reminder to change it

rre-dataset-generator/tests/integration/docker-compose.yml
nicolo-rinaldi: I would pick one way to do things and stick to that across all integration tests of all search engines. TBD

nicolo-rinaldi (commented):
I definitely like this library to set up integration tests, but we need to remember that we need to find a common way to set up folders. I think this is a key to improve the process of building integration tests for the search engines supported.

rre-dataset-generator/tests/integration/docker-compose.yml
Intrinsical-AI: I agree. This is exploratory for the reasearch

PR #175, opened 2025-07-22
DAGE-25: updated writers with Path input version

nicolo-rinaldi:
- added folders check for writers + Path-type support

### DIFF
--- a/rre-dataset-generator/src/writers/abstract_writer.py
+++ b/rre-dataset-generator/src/writers/abstract_writer.py
@@ -1,5 +1,6 @@
 from abc import ABC, abstractmethod
 from typing import List, Tuple
+from pathlib import Path
 
 from src.model.query_rating_context import QueryRatingContext
 from src.search_engine.data_store import DataStore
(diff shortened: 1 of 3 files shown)


Intrinsical-AI (dismissed):
Nice !

Intrinsical-AI (approved):

rre-dataset-generator/src/writers/abstract_writer.py
```diff
@@ -16,7 +17,7 @@ def __init__(self, datastore: DataStore):
         self.datastore = datastore
 
     @abstractmethod
-    def write(self, output_path: str) -> None:
+    def write(self, output_path: str | Path) -> None:
```
dantuzi, line 20: Just out of curiosity, why can't it be just `Path` instead of `str | Path`?

dantuzi (approved):

rre-dataset-generator/src/writers/abstract_writer.py
nicolo-rinaldi: To let this function be able to work also with `str` version of the Path. Can be changed also with just the Path version.

PR #176, opened 2025-07-22
DAGE-24: Feature/search engine uniquekey param

nicolo-rinaldi:
Add a REST call il Solr search engine interface to retrieve the unique key in the schema for the collection used to generate the dataset. Update also the search engine abstraction + test

[ticket](https://sease.atlassian.net/browse/DAGE-24)


rre-dataset-generator/src/search_engine/search_engine_base.py
```diff
@@ -31,6 +30,6 @@ def fetch_for_evaluation(self,
         pass
 
     @abstractmethod
-    def search(self, payload: Dict[str, Any], doc_fields: List[str]) -> List[Document]:
+    def search(self, payload: Dict[str, Any]) -> List[Document]:
```
Intrinsical-AI: this makes the trick to solve the current silent errors / lack of warning when asking for subset of fields, but If I understand correctly its a bit dirty: if the index is big; ej: each doc has many fields we might want to not load all - only the relevants for LLM generation, right?

An hybrid solution (even I recommend mandatory), could be making the field list optional
"def search(self, payload: Dict[str, Any], doc_fields: Optional[List[str]] = None) -> List[Document]:
"

rre-dataset-generator/src/search_engine/solr_search_engine.py
```diff
@@ -19,6 +19,7 @@ class SolrSearchEngine(BaseSearchEngine):
     def __init__(self, endpoint: str):
         super().__init__(endpoint)
         self.HEADERS = {'Content-Type': 'application/json'}
+        self.UNIQUE_KEY = requests.get(urljoin(endpoint, 'schema/uniquekey')).json()['uniqueKey']
```
Intrinsical-AI: In the same file, the function:

```
def fetch_for_query_generation(self,
                                   documents_filter: Union[None, List[Dict[str, List[str]]]],
                                   doc_number: int, doc_fields: List[str]) \
            -> List[Document]:
        payload = {
            'query': '*:*',
            'params': {
                'rows': doc_number,
                'fl' : doc_fields if 'id' in doc_fields else doc_fields + ['id']
            }
        }
```
        
I think it should be like:
`        'fl': doc_fields if self.UNIQUE_KEY in doc_fields else doc_fields + [self.UNIQUE_KEY]
`
        To keep consistant with the `UNIQUE_KEY` refactor

rre-dataset-generator/src/search_engine/search_engine_base.py
```diff
@@ -2,15 +2,14 @@
...
-
 class EndpointValidator(BaseModel):
     endpoint: HttpUrl
-
 class BaseSearchEngine(ABC):
     def __init__(self, endpoint: str):
         validated = EndpointValidator(endpoint=endpoint)
         self.endpoint = validated.endpoint
         self.PLACEHOLDER = "#$query##"
+        self.UNIQUE_KEY = 'id'
```
Intrinsical-AI, line 10: This assumes the Engine it's gonna focus on a single instance / schema. For example: with this pattern, we should use 2 instances for accessing the test index; (e.g:http://localhost:8983/solr/testcore/) and one additional for each real (e.g:http://localhost:8983/solr/prod_1/, e.g:http://localhost:8983/solr/prod_2/..) Could work for simpler approaches, but if we wnt to make it scalable, for example input a parameter into the constructo it's a better general solution.  E.g:

```
class SolrSearchEngine:
    def __init__(self, endpoint: str, core_name: str, unique_key: Optional[str] = None):
        self.core = core_name
        self.endpoint = f"{endpoint.rstrip('/')}/{core_name}/"
        self.UNIQUE_KEY = unique_key or self._resolve_unique_key()
```

Intrinsical-AI (commented):
I have some insights, would like to reach agreement about the approach

rre-dataset-generator/src/search_engine/search_engine_base.py
nicolo-rinaldi: In the same file, both the functions that call `search` function have the parameter `doc_fields` and deal with it depending on the use case. This choice has been done to keep the search function dependent ONLY on the payload, that must be built outside. I don't know if this makes sense for you. What do you think?

rre-dataset-generator/src/search_engine/search_engine_base.py
nicolo-rinaldi: I think is not necessary to deal with this function with `endpoint` and `core_name` param, since we would need to take the endpoint from the user and parse it. In my opinion, this is a feature we don't need now

rre-dataset-generator/src/search_engine/solr_search_engine.py
nicolo-rinaldi: Yes, you are absolutely right. Changed. Also in another similar line of code. Now it should be fine

rre-dataset-generator/src/search_engine/search_engine_base.py
Intrinsical-AI: Ok, fair. Just write a mental note on this in case we need to scale / change the approach

rre-dataset-generator/src/search_engine/solr_search_engine.py
```diff
@@ -110,10 +111,10 @@ def search(self, payload: Dict[str, Any]) -> List[Document]:
                 reformat_raw_doc = []
                 for doc in raw_docs:
                      clean_doc = dict()
-                     clean_doc['id'] = doc['id']
+                     clean_doc['id'] = doc[self.UNIQUE_KEY]
                      clean_doc['fields'] = dict()
                      for k, v in doc.items():
-                        if k != 'id':
+                        if k != self.UNIQUE_KEY:
```
Intrinsical-AI, line 119: two lines down: if v it's [] (pass the list check), isinstance(v[0,str) -> IndexError.

nicolo-rinaldi: can you elaborate a little bit better? do you want me to throw an IndexError if the list is empty?

rre-dataset-generator/src/search_engine/search_engine_base.py
Intrinsical-AI: The current design delegates to the payload the responsibility of including the `fl` parameter, but skipping it can have a big impact on memory and bandwidth.

For example: let’s suppose an index with documents of ~20 fields. If we only need `["id", "title", "body"]`, the average document size could be reduced from 10 KB to ~2.1 KB.

Hypothetical scenario:
- 100 documents per query (`rows=100`)
- 1000 queries (`R=1000`)

The overhead from not using `fl` would be:
((10 - 2.1) KB) × 100 documents × 1000 queries = ~768 MB

This implies a +367% increase in traffic and memory, **with no functional gain**.

-- Given this 

 I’d recommend making field filtering either:
- mandatory
- explicit via an optional parameter (`doc_fields: Optional[List[str]]`)

I think that's a big gain, especially when dealing with large indices

rre-dataset-generator/src/search_engine/solr_search_engine.py
Intrinsical-AI: It's gonna throw an unhadled errror when receving a empty list

nseidan (dismissed):

rre-dataset-generator/src/search_engine/solr_search_engine.py
nicolo-rinaldi: Now it should be fixed, take a look

rre-dataset-generator/src/search_engine/search_engine_base.py
nicolo-rinaldi: @dantuzi what do you think about this?

rre-dataset-generator/config.yaml
dantuzi, line 1: Is this a leftover of another task?

rre-dataset-generator/src/search_engine/solr_search_engine.py
```diff
@@ -16,9 +17,12 @@ class SolrSearchEngine(BaseSearchEngine):
...
-    def __init__(self, endpoint: str):
+    def __init__(self, endpoint: HttpUrl | str):
         super().__init__(endpoint)
         self.HEADERS = {'Content-Type': 'application/json'}
+        log.debug(f"Working on endpoint: {self.endpoint}")
+        self.UNIQUE_KEY = requests.get(urljoin(self.endpoint.encoded_string(), 'schema/uniquekey')).json()['uniqueKey']
+        log.debug(f"uniqueKey found: {self.UNIQUE_KEY}")
 
     @staticmethod
     def template_to_json_body(template_payload: str) -> Dict[str, Any]:
```
dantuzi: This function should return a list of Documents as it returns the result of the method `search`

rre-dataset-generator/src/search_engine/search_engine_base.py
dantuzi: Not a strong opinion on this. 
I believe that making field filtering mandatory may be too restrictive at this time.
Considering that `doc_fields` is a user parameter and not another random parameter, I like the idea of having it as the explicit and optional list.
On the other side, if we have the `doc_fields` parameter, we give the possibility to the method users to specify the parameter `fl` in the payload and in the dec_fields parameter, who wins?

That being said, the method `search` should be a "private" method, so:
- It should be renamed to `_search` (the underscore before the method name)
- I'm not even convinced we need this method in the super class
- This is not something blocking for now, and it can be tricky when we develop the client for OpenSearch, Elastic, or Vespa. In that case, we may need to change it. 

@nicolo-rinaldi feel free to implement the easiest solution for now

rre-dataset-generator/src/search_engine/solr_search_engine.py
```diff
@@ -16,9 +17,12 @@ class SolrSearchEngine(BaseSearchEngine):
...
     """
-    def __init__(self, endpoint: str):
+    def __init__(self, endpoint: HttpUrl | str):
         super().__init__(endpoint)
         self.HEADERS = {'Content-Type': 'application/json'}
+        log.debug(f"Working on endpoint: {self.endpoint}")
+        self.UNIQUE_KEY = requests.get(urljoin(self.endpoint.encoded_string(), 'schema/uniquekey')).json()['uniqueKey']
+        log.debug(f"uniqueKey found: {self.UNIQUE_KEY}")
 
     @staticmethod
```
dantuzi: I've just noticed, it shouldn't be static

rre-dataset-generator/tests/mocks/solr.py
dantuzi, line 1: What was the problem of having it in the same `test_solr_search_engine.py` file?

dantuzi (changes requested):

rre-dataset-generator/tests/mocks/solr.py
nicolo-rinaldi: With all mock classes inside that file it was becoming very long and difficult to read

rre-dataset-generator/src/search_engine/solr_search_engine.py
nicolo-rinaldi: removed

nicolo-rinaldi: and I made the method private

rre-dataset-generator/src/search_engine/solr_search_engine.py
nicolo-rinaldi: No, it doesn't. This function is just returning the payload. To avoid misunderstanding, the name has been changed to `_template_to_json_payload`

rre-dataset-generator/config.yaml
nicolo-rinaldi: yes

nicolo-rinaldi: I reverted it

rre-dataset-generator/src/search_engine/search_engine_base.py
nicolo-rinaldi: Changed into `_search`. I'll keep the current implementation, since it's faster. We have this github history if at some point we need to change into the other version. For me, the design of the payload must be handled outside the search method, so that the search function just performs the search, asking the search system.

rre-dataset-generator/src/search_engine/solr_search_engine.py
dantuzi: You're right, sorry. I got confused by GitHub because I didn't see some lines in the middle

dantuzi (dismissed):

dantuzi (approved):

PR #177, opened 2025-07-22
DAGE-25: v1 working demo

nicolo-rinaldi:
Full pipeline in the file `dataset_generator.py`. Number of total queries fixed. Implemented the score of the full cartesian product between retrieved documents and queries. 


rre-dataset-generator/dataset_generator.py
```diff
@@ -16,12 +21,88 @@
...
+                                                       doc_fields=config.doc_fields)
+        for doc in docs_eval:
+            if not data_store.has_document(doc.id):
+                data_store.add_document(doc.id, doc)
+            data_store.add_query(query_rating_context.get_query(), doc.id)
+
+    log.debug(f"Number of documents evaluated: {len(docs_to_generate_queries)}")
+    for query_rating_context in data_store.get_queries():
+        for doc in data_store.get_documents():
+            data_store.add_query(query_rating_context.get_query(), doc.id)
```
dantuzi: This is out of the scope of this ticket

rre-dataset-generator/dataset_generator.py
```diff
@@ -16,12 +21,88 @@
...
+            for line in file:
+                if line.strip():
+                    user_queries.append(line.strip())
+
+    # retrieval of the documents needed to generate the queries
+    docs_to_generate_queries = search_engine.fetch_for_query_generation(documents_filter=config.documents_filter,
+                                                                        doc_number=config.doc_number,
+                                                                        doc_fields=config.doc_fields)
+    log.debug(f"Number of documents retrieved for generation: {len(docs_to_generate_queries)}")
+    llm = build_openai(LLMConfig.load(config.llm_configuration_file))
```
dantuzi: How about if the user adds a parameter to work with Gemini? Do we need to change the code?

nicolo-rinaldi: Yes, we need add an `if` statement

rre-dataset-generator/dataset_generator.py
```diff
@@ -16,12 +21,88 @@
...
+                                            doc_id,
+                                            score)
+
+    if config.output_format == 'quepid':
+        writer = QuepidWriter(data_store)
+        writer.write(config.output_destination)
+    else:
+        error_msg = "Other format than 'quepid' are not yet supported"
+        log.error(error_msg)
+        raise NotImplementedError(error_msg)
```
dantuzi: I would have used the factory pattern: https://www.tutorialspoint.com/design_pattern/factory_pattern.htm

rre-dataset-generator/dataset_generator.py
dantuzi: Let's use the factory pattern instead. Something like
```
LLMService = Factory.build_llm_service(config)
```

rre-dataset-generator/dataset_generator.py
nicolo-rinaldi: Ok, we need to update the writer with a writer factory then

rre-dataset-generator/dataset_generator.py

rre-dataset-generator/dataset_generator.py
nicolo-rinaldi: removed

rre-dataset-generator/dataset_generator.py
nicolo-rinaldi: I implemented this

PR #179, opened 2025-07-23
Dage 18: Persistent Storage JSON

Intrinsical-AI:
### TASK TICKET 
 [DAGE-18](https://sease.atlassian.net/jira/software/projects/DAGE/boards/84?selectedIssue=DAGE-18) - Includes a mini-research on mocks vs real I/O for files). 
 
 
 #### Changes

 - Improved docstrings on previous code
 - Improved logs information
 - Fixed OLD tests ("is" replaced by "==")
  - Added fns:
```       
          save_queries_and_docs(self, filepath: str | Path) -> None:
          load_queries_and_docs(self, filepath: str | Path) -> None:
          save_rating_triples(self, filepath: str | Path) -> None:
          load_rating_triples(self, filepath: str | Path) -> None:
```
- Added new tests to `/tests/unit/test_data_store.py`


rre-dataset-generator/src/model/query_rating_context.py
```diff
@@ -12,10 +12,13 @@ class QueryRatingContext:
 
     DOC_NOT_RATED: int = -1  # doc is not yet rated
 
-    def __init__(self, query: str, doc_id: str):
+    def __init__(self, query: str, doc_id: str | None = None):
         self._id: str = str(uuid.uuid4())
         self._query: str = query
-        self._doc_id_to_rating_score: Dict[str, int] = {doc_id: self.DOC_NOT_RATED}
+        self._doc_id_to_rating_score: Dict[str, int] = {}
+        # HANDLING NONEs - thowed error in some tests
```
nseidan: typo -> throw*

rre-dataset-generator/src/search_engine/data_store.py
```diff
@@ -35,7 +36,7 @@ def _get_document(self, doc_id: str) -> Optional[Document]:
 
     def add_document(self, doc_id: str, document: Document) -> None:
         if doc_id in self._documents:
-            log.error("Document  id %s found in DataStore", doc_id)
+            log.error("Document id %s already exists in DataStore", doc_id)
```
nseidan: curious, why changed this sentence? found vs already exists?  it's the same.

rre-dataset-generator/src/search_engine/data_store.py
```diff
@@ -113,3 +114,74 @@ def has_rating_score(self, query_id: str, doc_id: str) -> bool:
...
 
+    
+    def save_queries_and_docs(self, filepath: str | Path) -> None:
+        """
+        Saves all query contexts (query_id, text, doc_ids) to a JSON file.
+        """
+        data = [
+            {
+                "query_id": ctx.get_query_id(),
+                "query_text": ctx.get_query(),
```
nseidan: let's keep the name `context` instead of `ctx`; avoid shortened naming in general; in the file, also already it is named as `context`.

rre-dataset-generator/src/search_engine/data_store.py
```diff
@@ -113,3 +114,74 @@ def has_rating_score(self, query_id: str, doc_id: str) -> bool:
...
+            self._queries_by_id[context.get_query_id()] = context
+            self._query_text_to_query_id[context.get_query()] = context.get_query_id()
+
+
+    def save_rating_triples(self, filepath: str | Path) -> None:
+        """
+        Saves all (query_id, doc_id, score) triples to a JSON file.
+        """
+        triples = []
+        for ctx in self._queries_by_id.values():
```
nseidan: similar, keep `context`, not `ctx`

rre-dataset-generator/src/search_engine/data_store.py
```diff
@@ -113,3 +114,74 @@ def has_rating_score(self, query_id: str, doc_id: str) -> bool:
...
+            qid = ctx.get_query_id()
+            for doc_id in ctx.get_doc_ids():
+                try:
+                    score = ctx.get_rating_score(doc_id)
+                    triples.append({
+                        "query_id": qid,
```
nseidan: keep it consistent:  somewhere in the code it's `qid`  or sometimes it's `query_id` for query id.

rre-dataset-generator/src/search_engine/data_store.py
```diff
@@ -113,3 +114,74 @@ def has_rating_score(self, query_id: str, doc_id: str) -> bool:
...
+
+    def load_rating_triples(self, filepath: str | Path) -> None:
+        """
+        Loads rating triples from a JSON file and updates existing QueryRatingContext entries.
+        """
+        with open(filepath, "r", encoding="utf-8") as f:
+            triples = json.load(f)
+
+        for triple in triples:
+            query_id = triple["query_id"]
```
nseidan: keep it consistent: somewhere in the code it's qid or sometimes it's query_id for query id.

rre-dataset-generator/src/search_engine/data_store.py
Intrinsical-AI: It's not the same. 

"Document  id %s found in DataStore" It's a generic error, that could be assumed associated with a search function.

 In this case we're logging an error when we try to add an already existing doc into the memory structure, so the best option it's to explicit that the associated ID is already inserted.
 
 
 From my perspective even better description would be: "Detected an error when adding document to the data store. Document {doc_id} already present."
 
  -> It's descriptive at a high level on the failure

rre-dataset-generator/src/search_engine/data_store.py
Intrinsical-AI: I accept your point, but I don't fully agree: making variable names to "more" descriptive / verbose, does not always equal to  "best practices". 

For example: in local contexts (especially in comprehensive lists), one python convention (therefore, the path to follow) its to: 
"Shorten local variable names for brevity and scope-limiting" - Source: https://llego.dev/posts/rules-variable-naming-identifiers-python/

That especially applies on comprehensive lists and/or loop variables.

rre-dataset-generator/src/search_engine/data_store.py
Intrinsical-AI: Responded in the other comment. Agree here makes more sense to rename to "_context".
_ it's used for naming local / private variables or names. I think fits here.

rre-dataset-generator/src/search_engine/data_store.py
Intrinsical-AI: As mentioned earlier:

" in local contexts (especially in comprehensive lists), one python convention (therefore, the path to follow) its to:
"Shorten local variable names for brevity and scope-limiting" - Source: https://llego.dev/posts/rules-variable-naming-identifiers-python/"

Here the variable it's the core of the query, so makes sense to keep it verbose. When using qid, it's probably relative to temporal / local / loop vars

rre-dataset-generator/src/search_engine/data_store.py
Intrinsical-AI: Here makes sense to update to query_id. More explicit

rre-dataset-generator/src/model/query_rating_context.py
Intrinsical-AI: Thanks for the spot! Comming in the next commit

nseidan: Please resolve when you commit the change.

rre-dataset-generator/src/search_engine/data_store.py
nseidan: fine with either context or _context, just to have more descriptive naming. 
Feel free to resolve  this when you update it.

rre-dataset-generator/src/search_engine/data_store.py
nseidan: Feel free to resolve this when you update it.

rre-dataset-generator/src/search_engine/data_store.py
nseidan: it doesn't matter if the var is used locally/temp etc , it should be maintainable and readable for others and that's the point.

nseidan (dismissed):

nseidan:
I approved this, but please make changes on the things mentioned and feel free to resolve the conversations once you update .

rre-dataset-generator/src/model/query_rating_context.py
Intrinsical-AI: Update: this should be solved

rre-dataset-generator/src/search_engine/data_store.py
Intrinsical-AI: Update:
I still thinking here makes sense to shorten, since its 1) local 2) a loop 3) the description itself talks about context

Let the rest of the team decide

rre-dataset-generator/src/search_engine/data_store.py
Intrinsical-AI: Update:
Used _context for redability. As it's a loop could be also ctx, but I ackowledge there's a wide range of opinions, and redability matters, so I agree / makes sense for me to change,

rre-dataset-generator/src/search_engine/data_store.py
Intrinsical-AI: Changed to query_id for redability / consistency

rre-dataset-generator/src/search_engine/data_store.py
Intrinsical-AI: I agree clarity is key, but just as reminder: Python's own naming guidelines recommend shorter names for local/temporary variables, especially in comprehensions.

See: PEP8 and common practice in modern libraries like LangChain or Haystack. The community, in the reference libraries, often use shortened local variables like ctx, especially in comprehensions or loops.

image

[Source](https://docs.smith.langchain.com/reference/python/_modules/langsmith/run_helpers)

rre-dataset-generator/src/search_engine/data_store.py
dantuzi: Folks, we are talking about a variable name used in three lines of code.
I prefer the name `context`, but I'm okay with leaving the compacted version in the list comprehension.
@Intrinsical-AI, as a general rule, I'd prefer meaningful names, but in this specific and restricted case, feel free to do what you prefer

rre-dataset-generator/src/search_engine/data_store.py
dantuzi: I quickly read the comments but it seems the same discussion of a conversation above. Please don't waste time on a small name

rre-dataset-generator/src/model/query_rating_context.py
```diff
@@ -12,10 +12,13 @@ class QueryRatingContext:
...
 
-    def __init__(self, query: str, doc_id: str):
+    def __init__(self, query: str, doc_id: str | None = None):
         self._id: str = str(uuid.uuid4())
         self._query: str = query
-        self._doc_id_to_rating_score: Dict[str, int] = {doc_id: self.DOC_NOT_RATED}
+        self._doc_id_to_rating_score: Dict[str, int] = {}
+        # HANDLING NONEs - throwed error in some tests
+        if doc_id is not None:
+            self._doc_id_to_rating_score[doc_id] = self.DOC_NOT_RATED
```
nicolo-rinaldi, line 23: this is very useful also in the demo to have the possibility to add a query without the doc.id.

nicolo-rinaldi (dismissed):

rre-dataset-generator/src/search_engine/data_store.py
```diff
@@ -1,6 +1,8 @@
 from __future__ import annotations
 
+import json
 import logging
+from pathlib import Path
 from typing import Dict, List, Optional
```
dantuzi, line 8: add a constant or a env variable to specify where to store and load the file

how about "tmp/datastore.json"

rre-dataset-generator/src/search_engine/data_store.py
```diff
@@ -10,33 +12,35 @@
...
 
-
 class DataStore:
     """
     Stores/retrieves documents, queries, and rating scores.
     """
 
     def __init__(self):
-        self._documents: Dict[str, Document] = {}  # doc_id -> Document
+        self._documents: Dict[str, Document] = {}                # doc_id -> Document
```
dantuzi: We should load the old partial data if they exists

rre-dataset-generator/src/search_engine/data_store.py
Intrinsical-AI: I agree :+1: 
The tmp file it's now created in the rre-dataset-generator dir

rre-dataset-generator/src/search_engine/data_store.py
Intrinsical-AI: Implemented in the last commit

Intrinsical-AI:
@dantuzi should make more sense in general now

rre-dataset-generator/src/search_engine/data_store.py
```diff
@@ -112,3 +125,77 @@ def has_rating_score(self, query_id: str, doc_id: str) -> bool:
         context = self._get_query_rating_context_by_id(query_id)
         return context.has_rating_score(doc_id)
 
+    
+    @staticmethod
+    def query_context_docs_to_dict(query: QueryRatingContext, documents: List[Document]) -> dict:
+        return {
+            "query_id": query.get_query_id(),
+            "query_text": query.get_query(),
+            "doc_ids": query.get_doc_ids(),
```
dantuzi: `query.get_doc_ids()` returns the list of keys of `query._doc_id_to_rating_score`.
Why do we need to keep both `doc_ids` and `doc_ratings` if `doc_ratings` is the superset of `doc_ids`?

rre-dataset-generator/src/search_engine/data_store.py
```diff
@@ -112,3 +125,77 @@ def has_rating_score(self, query_id: str, doc_id: str) -> bool:
...
+            "doc_ratings": query._doc_id_to_rating_score.copy(),  # copy to avoid mutation
+            "documents": [doc.model_dump() for doc in documents if doc is not None]
```
dantuzi: what does `model_dump()` do?

rre-dataset-generator/src/search_engine/data_store.py
```diff
@@ -112,3 +125,77 @@ def has_rating_score(self, query_id: str, doc_id: str) -> bool:
...
+        """
+        global TMP_FILE
+        if filepath is None:
+            filepath = TMP_FILE
+
+        all_content = []
+        for query_ctx in self._queries_by_id.values():
+            docs_ = [self.get_document(doc_id) for doc_id in query_ctx.get_doc_ids()]
+            data = self.query_context_docs_to_dict(query_ctx, docs_)
+            all_content.append(data)
```
dantuzi: Just out of curiosity, if you do `json.dump(self._queries_by_id)`, does it work?

I see that your solution also stores the document texts.

rre-dataset-generator/src/search_engine/data_store.py
Intrinsical-AI: Doc it's a Pydantic model. model_dump it's the internal fn to serialize (to dict / json)

rre-dataset-generator/src/search_engine/data_store.py
Intrinsical-AI: Good question, you're right, I re-checked and it's a bit redundant. Tthere's a simpler approach: _doc_id_to_rating_score is a {doc_id: int_rating}, so we could just use list(_doc_id_to_rating_score.keys()) to get the IDs for example.

The only downside I see is that dicts don’t keep the order, so now that I think about it, keeping them as it is might help ensure consistency across runs if needed

rre-dataset-generator/src/search_engine/data_store.py
Intrinsical-AI: If the content it's not an object, should work.  From the repository json
```
def dumps(obj, *, skipkeys=False, ensure_ascii=True, check_circular=True,
        allow_nan=True, cls=None, indent=None, separators=None,
        default=None, sort_keys=False, **kw):
    """Serialize ``obj`` to a JSON formatted ``str``.
    If ``skipkeys`` is true then ``dict`` keys that are not basic types
    (``str``, ``int``, ``float``, ``bool``, ``None``) will be skipped
    instead of raising a ``TypeError``.....""""
```
But in this case self._queries_by_id: Dict[str, QueryRatingContext] = {} and QueryRatingContext it's not pydantic (we don't have model_dumps()), neither implement a serialization method. 

So it doesn't work with the current setup but we could implement serialization with __dict__ for example to allow casting - `dict(query)`

rre-dataset-generator/src/search_engine/data_store.py
Intrinsical-AI: @dantuzi Update: I was wrong. Dictionaries nowadays supports order (python >3.7) -> consistant between runs if inputs are deterministic -> tackle the previous point.

Also: reflecting; seems better in general not having to update 2 distinct fields. 

Will upload the updated code tomorrow following your review. Thanks!
 :+1:

rre-dataset-generator/src/search_engine/data_store.py
Intrinsical-AI: Mini PoC

image

image

Intrinsical-AI:
###  feat: datastore: JSON persistence + QueryContext API cleanup

Added on-disk persistence (`./tmp/datastore.json`) to make runs reproducible. Cleaned up `QueryRatingContext` with proper (de)serialization and safer access to ratings.

**Included**

* `DataStore.save/load_tmp_file_content()` with custom serializer.
* Stable IDs in `QueryRatingContext`; `.to_dict()` / `.from_dict()` methods.
* `get_query()` → `get_query_text()`; `get_rating_score()` raises if missing.
* Field validator in `Document` to reject empty fields.
* Writers/tests updated to use the new API; TMP file path patchable for tests.

rre-dataset-generator/tests/unit/test_data_store.py
```diff
@@ -1,53 +1,123 @@
...
+        ds.load_tmp_file_content()
 
 
+@pytest.mark.parametrize("content", [
+    # Mismatch outer (key) vs inner query_id -> debe fallar
+    {
+        "queries": {"q1": {"query_id": "DIFF", "query_text": "x", "doc_ratings": {}}},
+        "documents": {},
+    },
+    # Documento con id interno distinto al de la clave -> debe fallar
```
dantuzi: comment in spanish

rre-dataset-generator/tests/unit/test_query_rating_context.py
```diff
@@ -0,0 +1,51 @@
...
+    ctx.add_doc_id("d1")
+    ctx.add_doc_id("d1")  # should not raise or duplicate
+    assert ctx.get_doc_ids() == ["d1"]
+
+
+def test_missing_rating_raises_keyerror():
+    ctx = QueryRatingContext(query="AI")
+    ctx.add_doc_id("d1")
+    with pytest.raises(KeyError):
+        ctx.get_rating_score("d1")
```
dantuzi: new line

rre-dataset-generator/tests/unit/test_data_store.py
```diff
@@ -1,53 +1,123 @@
...
+
+@pytest.fixture
+def set_tmp_file(tmp_path, monkeypatch):
+    """Fija la ruta del datastore a un archivo temporal y la devuelve."""
+    path = tmp_path / "datastore.json"
+    monkeypatch.setattr(data_store, "TMP_FILE", str(path), raising=False)
+    return path
+
+
+def test_add_and_get_document(empty_store):
```
dantuzi: can we use the naming convention test_<what we are testing>__expect__<what to expect>

rre-dataset-generator/tests/unit/test_query_rating_context.py
```diff
@@ -0,0 +1,51 @@
+# tests/test_query_rating_context.py
+
+import pytest
+from src.model.query_rating_context import QueryRatingContext
+
+
+def test_initialization_with_doc_id():
```
dantuzi: the name convention for tests

rre-dataset-generator/tests/unit/test_data_store.py
Intrinsical-AI: Fixed (all across)

rre-dataset-generator/tests/unit/test_data_store.py
Intrinsical-AI: Added naming conventions

rre-dataset-generator/tests/unit/test_query_rating_context.py
Intrinsical-AI: Added

dantuzi (approved):

dantuzi (approved):

PR #180, opened 2025-07-23
DAGE-25: main demo v1

nicolo-rinaldi:
Full pipeline in the file dataset_generator.py. Number of total queries fixed. Implemented the score of the full cartesian product between retrieved documents and queries.


rre-dataset-generator/dataset_generator.py
```diff
@@ -1,27 +1,85 @@
...
+                break
+            query_id = data_store.add_query(query_text, doc.id)
+            data_store.add_rating_score(query_id, doc.id, 2)
+        if flag:
+            break
+
+    log.debug(f"Number of documents evaluated: {len(docs_to_generate_queries)}")
+    for query_rating_context in data_store.get_queries():
+        for doc in data_store.get_documents():
+            data_store.add_query(query_rating_context.get_query(), doc.id)
```
nseidan: adding new queries to the data_store while iterating over ` data_store.get_queries()` ?  -> you might have subtle bugs there ? it's better to store `data_store.get_queries() `in a local var then iterate over this.

rre-dataset-generator/dataset_generator.py
```diff
@@ -1,27 +1,85 @@
...
         configure_logging(logging.INFO)
+    log = logging.getLogger(__name__)
+
+    search_engine = build_search_engine(search_engine_type=config.search_engine_type,
+                                        endpoint=config.search_engine_collection_endpoint)
+    data_store = DataStore()
+
+    user_queries = []
+    if config.queries is not None:
+        with open(config.queries, 'r', encoding='utf-8') as file:
```
nseidan, line 40: is the file valid/already validated? no try/catch ?

rre-dataset-generator/dataset_generator.py
```diff
@@ -1,27 +1,85 @@
...
+    flag = False
+    for doc in docs_to_generate_queries:
+        data_store.add_document(doc.id, doc)
+        query_texts = service.generate_queries(doc, num_queries_per_doc)
+        for query_text in query_texts:
+            if len(data_store.get_queries()) >= config.num_queries_needed:
+                flag = True
+                break
+            query_id = data_store.add_query(query_text, doc.id)
+            data_store.add_rating_score(query_id, doc.id, 2)
```
nseidan: assuming score type is graded not binary ?  because i see 2.

rre-dataset-generator/dataset_generator.py
```diff
@@ -1,27 +1,85 @@
...
+
+    # retrieval of the documents needed to generate the queries
+    docs_to_generate_queries = search_engine.fetch_for_query_generation(documents_filter=config.documents_filter,
+                                                                        doc_number=config.doc_number,
+                                                                        doc_fields=config.doc_fields)
+    log.debug(f"Number of documents retrieved for generation: {len(docs_to_generate_queries)}")
+    llm = build_chat_model(LLMConfig.load(config.llm_configuration_file))
+    service = LLMService(chat_model=llm)
+
+    num_queries_per_doc = int(( (config.num_queries_needed - len(user_queries)) // config.doc_number) * 1.5)
```
nseidan: btw if config .doc_number is 0, it will crash.

rre-dataset-generator/dataset_generator.py
nicolo-rinaldi: Yes, if the file is not there, pydantic's `Config` model is going to throw an error.

rre-dataset-generator/dataset_generator.py
nicolo-rinaldi: Pydantic's `Config` model is going to throw an error if it's not greater than 0

rre-dataset-generator/dataset_generator.py
nicolo-rinaldi: yes, you are absolutely right. Fixed

rre-dataset-generator/dataset_generator.py
nicolo-rinaldi: fixed

nseidan (dismissed):

rre-dataset-generator/src/search_engine/__init__.py
```diff
@@ -0,0 +1,18 @@
...
+from .solr_search_engine import SolrSearchEngine
+import logging
+
+log = logging.getLogger(__name__)
+
+SEARCH_ENGINE_REGISTRY = {
+    "solr": SolrSearchEngine,
+}
+
+def build_search_engine(search_engine_type: str, endpoint: HttpUrl) -> BaseSearchEngine:
```
dantuzi: I'm expecting the factory as a dedicated class `SearchEngineFactory` with a static method `build(search_engine_type: str, endpoint: HttpUrl)`.

If I have to be completely honest, I don't like this method in the __init__.py file

rre-dataset-generator/src/llm/llm_service.py
```diff
@@ -22,7 +23,7 @@ def generate_queries(self, document: Document, num_queries_generate_per_doc: int
         """
         system_prompt = (
             f"You are a helpful assistant! Generate {num_queries_generate_per_doc} "
-            "keyword-phrase-based queries based on the given document below. "
+            "queries based on the given document below. " #keyword-phrase-based
```
dantuzi: I guess we can directly remove the comment after the prompt

rre-dataset-generator/dataset_generator.py
```diff
@@ -1,27 +1,86 @@
...
 
     config = Config.load(args.config_file)
+
     if args.verbose:
         configure_logging(logging.DEBUG)
     else:
         configure_logging(logging.INFO)
+    log = logging.getLogger(__name__)
+
+    search_engine = build_search_engine(search_engine_type=config.search_engine_type,
```
dantuzi: I was expecting something like
```
search_engine: BaseSearchEngine = SearchEngineFactory.build(
   search_engine_type = config.search_engine_type,
   endpoint=config.search_engine_collection_endpoint
)
```

or

```
search_engine: BaseSearchEngine = SearchEngineFactory.build(config=config)
```

rre-dataset-generator/dataset_generator.py
```diff
@@ -1,27 +1,86 @@
...
+    llm = build_chat_model(LLMConfig.load(config.llm_configuration_file))
+    service = LLMService(chat_model=llm)
+
+    num_queries_per_doc = int(( (config.num_queries_needed - len(user_queries)) // config.doc_number) * 1.5)
+
+    # query generation step
+    flag = False
+    for doc in docs_to_generate_queries:
+        data_store.add_document(doc.id, doc)
+        query_texts = service.generate_queries(doc, num_queries_per_doc)
```
dantuzi: the name `query_texts` can be replaced with a simpler `queries`

rre-dataset-generator/dataset_generator.py
```diff
@@ -1,27 +1,86 @@
...
+                                                                        doc_number=config.doc_number,
+                                                                        doc_fields=config.doc_fields)
+    log.debug(f"Number of documents retrieved for generation: {len(docs_to_generate_queries)}")
+    llm = build_chat_model(LLMConfig.load(config.llm_configuration_file))
+    service = LLMService(chat_model=llm)
+
+    num_queries_per_doc = int(( (config.num_queries_needed - len(user_queries)) // config.doc_number) * 1.5)
+
+    # query generation step
+    flag = False
```
dantuzi: The name `flag` is not meaningful; I would have preferred something like `all_queries_generated`.

Alternatively, a probably cleaner solution is to write a specific function that returns when all queries have been generated

rre-dataset-generator/dataset_generator.py
```diff
@@ -1,27 +1,86 @@
...
+    search_engine = build_search_engine(search_engine_type=config.search_engine_type,
+                                        endpoint=config.search_engine_collection_endpoint)
+    data_store = DataStore()
+
+    user_queries = []
+    if config.queries is not None:
+        with open(config.queries, 'r', encoding='utf-8') as file:
+            for line in file:
+                if line.strip():
+                    user_queries.append(line.strip())
```
dantuzi: We are adding all queries provided by the user into the variable `user_queries`, and we never read from it.
I suggest directly adding all user queries to the datastore (we don't specify the docId, so it probably requires updating the datastore), so we've already have them ready when we calculate the cartesian product

rre-dataset-generator/dataset_generator.py
```diff
@@ -1,27 +1,86 @@
...
+            query_id = data_store.add_query(query_text, doc.id)
+            data_store.add_rating_score(query_id, doc.id, max(config.relevance_label_set))
+        if flag:
+            break
+
+    log.debug(f"Number of documents evaluated: {len(docs_to_generate_queries)}")
+    queries_to_add = data_store.get_queries()
+    for query_rating_context in queries_to_add:
+        for doc in data_store.get_documents():
+            data_store.add_query(query_rating_context.get_query(), doc.id)
```
dantuzi: I don't understand why you need this

rre-dataset-generator/dataset_generator.py
nicolo-rinaldi: changed the name of the flag

rre-dataset-generator/dataset_generator.py
nicolo-rinaldi: changed

rre-dataset-generator/src/llm/llm_service.py
nicolo-rinaldi: removed

rre-dataset-generator/src/search_engine/__init__.py
nicolo-rinaldi: refactored and changed the file of all 3 factories

rre-dataset-generator/dataset_generator.py

rre-dataset-generator/dataset_generator.py
nicolo-rinaldi: To do the cartesian product and match every query with every document

dantuzi (approved):

rre-dataset-generator/dataset_generator.py
dantuzi: changed in the new version by @nicolo-rinaldi

PR #181, opened 2025-07-25
DAGE-17: added the query template

nicolo-rinaldi:
- added the few lines of code to retrieve documents given the queries and added to the datastore

[ticket](https://sease.atlassian.net/browse/DAGE-17)

### DIFF
--- a/rre-dataset-generator/config.yaml
+++ b/rre-dataset-generator/config.yaml
@@ -2,7 +2,7 @@
 # dataset-generator module
 # Default: "q=#$query##"
 # Example (solr):
-query_template: "q=#$query##&df=description&rows=3&q.op=OR&wt=json"
+query_template: "q=#$query##&df=description&rows=2&q.op=OR&wt=json"
 
 # The type of search engine to use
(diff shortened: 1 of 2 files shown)


rre-dataset-generator/config.yaml
```diff
@@ -2,7 +2,7 @@
 # dataset-generator module
 # Default: "q=#$query##"
 # Example (solr):
-query_template: "q=#$query##&df=description&rows=3&q.op=OR&wt=json"
+query_template: "q=#$query##&df=description&rows=2&q.op=OR&wt=json"
```
Intrinsical-AI, line 5: Just curious: this is to get 2 results instead of one? Each row one doc?

Intrinsical-AI (approved):
🏄

rre-dataset-generator/config.yaml
nicolo-rinaldi: Nope, this is to have 2 docs each query that we have after generation instead of 3. To reduce the number of API calls, since this is a demo. The cartesian product makes the number of calls explode

dantuzi (approved):

PR #183, opened 2025-07-25
Adding configuration to run the unit tests automatically in each PR

dantuzi:
### Notes
Adding the configuration to let GitHub run the unit tests automatically on each PR, so we avoid committing code that breaks the unit tests.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -1,0 +1,31 @@
+name: Unit Tests
+
+on:
+  pull_request:
+    branches: [ main, dataset-generator ]
+
+jobs:
+  test:
+    runs-on: ubuntu-latest
+
+    steps:
+    - name: Checkout code
+      uses: actions/checkout@v4
+
+    - name: Set up Python
+      uses: actions/setup-python@v4
+      with:
+        python-version: "3.11"
+
+    - name: Install uv
+      uses: astral-sh/setup-uv@v3
+      with:
+        version: "latest"
+
+    - name: Install dependencies
+      run: uv sync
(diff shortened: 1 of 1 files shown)


.github/workflows/test.yml
```diff
@@ -0,0 +1,29 @@
...
+    - name: Install uv
+      uses: astral-sh/setup-uv@v3
+      with:
+        version: "latest"
+
+    - name: Install dependencies
+      run: uv sync
+
+    - name: Run unit tests
+      run: uv run pytest
```
Intrinsical-AI, line 30: One question: it's worth including the integration tests? Now we're not yet implemented pytest-docker, so if intended I think we would have to manually run the integration tests for Solr (and the incoming OpenSearch / Vespa), maybe works only running the bash scripts ? (@nseidan included healthy checks and data loading, I'm copying them) 

Maybe integration tests are not needed yet

Intrinsical-AI (dismissed):

Intrinsical-AI:
Btw I approved but just realized that the test are not pasing 

```
Run uv sync
error: No `pyproject.toml` found in current directory or any parent directory
Error: Process completed with exit code 2.
```

I think shoud cd rre-dataset-generator/ or smth like that

.github/workflows/test.yml
dantuzi: 1. We still don't have integration tests
2. Integration tests have to work with a real instance of Solr (dockerizable) and LLM (I still have to read your document, so I don't know yet).

I believe that automating the unit test is still better than nothing

nicolo-rinaldi (approved):

PR #184, opened 2025-07-25
DAGE-19: Add OpenSearch support

nseidan:
Jira Issue: https://sease.atlassian.net/browse/DAGE-19 

* Added OpenSearch test env (`docker-compose.opensearch.yml`, `opensearch-init`)
* Added `OpenSearchEngine` to handle search 
* Added a unit test

Tested manually, and it is working as expected.

### DIFF
--- a/rre-dataset-generator/config.yaml
+++ b/rre-dataset-generator/config.yaml
@@ -9,6 +9,7 @@ query_template: "q=#$query##&df=description&rows=3&q.op=OR&wt=json"
 search_engine_type: "solr"
 
 # Endpoint URL of the search engine collection or index
+# opensearch: http://localhost:9200/testcore/
 search_engine_collection_endpoint: "http://localhost:8983/solr/testcore/"
 
 # (Optional) Filter query to restrict the set of documents used to generate queries.
--- a/rre-dataset-generator/src/search_engine/opensearch_engine.py
+++ b/rre-dataset-generator/src/search_engine/opensearch_engine.py
@@ -1,3 +1,4 @@
+import json
 import logging
 from typing import List, Dict, Any, Union
 
@@ -99,15 +100,27 @@ class OpenSearchEngine(BaseSearchEngine):
         return result
 
     @staticmethod
-    def _normalize(value: Any) -> Any:
-        """
-        Normalize a field value:
-        - string → return [cleaned string]
-        - list of strings → return [cleaned strings]
-        - else → return as it is
-        """
-        if isinstance(value, str):
-            return [clean_text(value)]
-        if isinstance(value, list) and all(isinstance(i, str) for i in value):
-            return [clean_text(i) for i in value]
-        return value
+    def _normalize(value: Any) -> List[str]:
+        """Normalize a field value into a list of cleaned strings or throws an exception."""
+        try:
+            if value is None:
+                return []
+
+            if isinstance(value, str):
+                return [clean_text(value)]
+
+            if isinstance(value, list):
+                return [clean_text(v) if isinstance(v, str) else str(v) for v in value]
+
+            if isinstance(value, dict):
+                cleaned_dict = {
+                    k: clean_text(v) if isinstance(v, str) else v
+                    for k, v in value.items()
+                }
+                return [json.dumps(cleaned_dict)]
+
+            return [str(value)]
+
+        except Exception as e:
+            raise ValueError(f"Failed to normalize value: {value}") from e
+
--- a/rre-dataset-generator/src/search_engine/opensearch_engine.py
+++ b/rre-dataset-generator/src/search_engine/opensearch_engine.py
@@ -1,0 +1,113 @@
+import logging
+from typing import List, Dict, Any, Union
+
+import requests
+from pydantic import HttpUrl
+from requests.exceptions import HTTPError, ConnectionError, Timeout, RequestException
+
+from src.model.document import Document
+from src.search_engine.search_engine_base import BaseSearchEngine
+from src.utils import clean_text
+
+log = logging.getLogger(__name__)
+
+
+class OpenSearchEngine(BaseSearchEngine):
+    """
+    OpenSearch implementation to search in a given index.
+    """
+
+    def __init__(self, endpoint: HttpUrl):
+        super().__init__(endpoint)
+        self.HEADERS = {'Content-Type': 'application/json'}
+        self.UNIQUE_KEY = "id"
+
+    def fetch_for_query_generation(self,
+                                   documents_filter: Union[None, List[Dict[str, List[str]]]],
+                                   doc_number: int,
+                                   doc_fields: List[str]) -> List[Document]:
+        filters = []
+        if documents_filter:
+            for field_values in documents_filter:
+                for field, values in field_values.items():
+                    if not values:
+                        continue
+                    if len(values) == 1:
+                        filters.append({"term": {field: values[0]}})
+                    else:
+                        filters.append({"terms": {field: values}})
+
+        fields = doc_fields if self.UNIQUE_KEY in doc_fields else doc_fields + [self.UNIQUE_KEY]
+
+        if filters:
+            query = {
+                "bool": {
+                    "filter": filters
+                }
+            }
+        else:
+            query = {
+                "match_all": {}
+            }
+
+        payload = {
+            "query": query,
+            "_source": fields,
+            "size": doc_number
+        }
+
+        return self._search(payload)
+
+    def fetch_for_evaluation(self, query_template: str, doc_fields: List[str], keyword: str = "*") -> List[Document]:
+        query = query_template.replace(self.PLACEHOLDER, keyword)
+        fields = doc_fields if self.UNIQUE_KEY in doc_fields else doc_fields + [self.UNIQUE_KEY]
+
+        payload = {
+            "query": {
+                "query_string": {
+                    "query": query
+                }
+            },
+            "_source": fields
+        }
+        return self._search(payload)
+
+    def _search(self, payload: Dict[str, Any]) -> List[Document]:
+        search_url = f"{self.endpoint}/_search"
+        try:
+            response = requests.post(search_url, headers=self.HEADERS, json=payload)
+            response.raise_for_status()
+        except (ConnectionError, Timeout, RequestException, HTTPError) as e:
+            log.error(f"OpenSearch query failed: {e}\nPayload: {payload}")
+            raise
+
+        hits = response.json().get("hits", {}).get("hits", [])
+        result = []
+
+        for hit in hits:
+            source = hit.get("_source", {})
+            doc_id = source.get("id", hit.get("_id"))
+
+            fields = {
+                key: self._normalize(value)
+                for key, value in source.items()
+                if key != "id"
+            }
+
+            result.append(Document(id=doc_id, fields=fields))
+
+        return result
+
+    @staticmethod
+    def _normalize(value: Any) -> Any:
+        """
+        Normalize a field value:
+        - string → return [cleaned string]
+        - list of strings → return [cleaned strings]
+        - else → return as it is
+        """
+        if isinstance(value, str):
+            return [clean_text(value)]
+        if isinstance(value, list) and all(isinstance(i, str) for i in value):
+            return [clean_text(i) for i in value]
+        return value
--- a/rre-dataset-generator/src/search_engine/opensearch_engine.py
+++ b/rre-dataset-generator/src/search_engine/opensearch_engine.py
@@ -27,6 +27,7 @@ class OpenSearchEngine(BaseSearchEngine):
                                    documents_filter: Union[None, List[Dict[str, List[str]]]],
                                    doc_number: int,
                                    doc_fields: List[str]) -> List[Document]:
+        """Fetches a list of documents for query generation based on optional filters."""
         filters = []
         if documents_filter:
             for field_values in documents_filter:
@@ -60,6 +61,7 @@ class OpenSearchEngine(BaseSearchEngine):
         return self._search(payload)
 
     def fetch_for_evaluation(self, query_template: str, doc_fields: List[str], keyword: str = "*") -> List[Document]:
+        """Fetches documents for evaluation by executing a query built from a template."""
         query = query_template.replace(self.PLACEHOLDER, keyword)
         fields = doc_fields if self.UNIQUE_KEY in doc_fields else doc_fields + [self.UNIQUE_KEY]
 
@@ -74,7 +76,9 @@ class OpenSearchEngine(BaseSearchEngine):
         return self._search(payload)
 
     def _search(self, payload: Dict[str, Any]) -> List[Document]:
+        """Perform a search to OpenSearch and return matching documents based on the given payload."""
         search_url = f"{self.endpoint}/_search"
+        log.debug(f"User-specified fields: {payload.get('_source')}")
         try:
             response = requests.post(search_url, headers=self.HEADERS, json=payload)
             response.raise_for_status()
@@ -87,6 +91,7 @@ class OpenSearchEngine(BaseSearchEngine):
 
         for hit in hits:
             source = hit.get("_source", {})
+            log.debug(f"Opensearch returns fields based on payload: {list(source.items())}")
             doc_id = source.get("id", hit.get("_id"))
 
             fields = {
--- a/rre-dataset-generator/tests/integration/docker-compose.opensearch.yml
+++ b/rre-dataset-generator/tests/integration/docker-compose.opensearch.yml
@@ -1,0 +1,33 @@
+services:
+  opensearch:
+    image: opensearchproject/opensearch:3.1.0
+    container_name: opensearch
+    environment:
+      - discovery.type=single-node
+      - node.name=opensearch
+      - bootstrap.memory_lock=true
+      - DISABLE_INSTALL_DEMO_CONFIG=true
+      - DISABLE_SECURITY_PLUGIN=true
+    ports:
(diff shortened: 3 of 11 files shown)


rre-dataset-generator/config.yaml
```diff
@@ -9,6 +9,7 @@ query_template: "q=#$query##&df=description&rows=3&q.op=OR&wt=json"
 search_engine_type: "solr"
 
 # Endpoint URL of the search engine collection or index
+# opensearch: http://localhost:9200/testcore/
```
nicolo-rinaldi, line 12: do we want to add this comment?

rre-dataset-generator/tests/integration/docker-compose.opensearch.yml
```diff
@@ -0,0 +1,33 @@
...
+      - discovery.type=single-node
+      - node.name=opensearch
+      - bootstrap.memory_lock=true
+      - DISABLE_INSTALL_DEMO_CONFIG=true
+      - DISABLE_SECURITY_PLUGIN=true
+    ports:
+      - 9200:9200
+      - 9600:9600
+
+  opensearch-dashboards:
```
nicolo-rinaldi, line 15: Is this needed?

rre-dataset-generator/tests/unit/resources/good_config_opensearch.yaml
```diff
@@ -0,0 +1,14 @@
+query_template: "q=#$query## AND title:cars"
```
nicolo-rinaldi, line 1: In elasticsearch I dealt with this template in a different way, so using the json-like query. What do you think is the best approach?

nicolo-rinaldi (dismissed):

rre-dataset-generator/config.yaml
nseidan: just added here as well as wanted to have other two: elastic and vespa, in particular port numbers and solr can stay as default.  We could also remove it.

rre-dataset-generator/tests/integration/docker-compose.opensearch.yml
nseidan: it is for manual testing purpose, to use dev tools for querying index.

rre-dataset-generator/tests/unit/resources/good_config_opensearch.yaml
nseidan: Opensearch handles json part out of the box, so thus provided a line string, but for elasticsearch i am not sure, never worked with. And to your question, no clue.

rre-dataset-generator/src/search_engine/opensearch_engine.py
```diff
@@ -0,0 +1,113 @@
...
+
+    def __init__(self, endpoint: HttpUrl):
+        super().__init__(endpoint)
+        self.HEADERS = {'Content-Type': 'application/json'}
+        self.UNIQUE_KEY = "id"
+
+    def fetch_for_query_generation(self,
+                                   documents_filter: Union[None, List[Dict[str, List[str]]]],
+                                   doc_number: int,
+                                   doc_fields: List[str]) -> List[Document]:
```
Intrinsical-AI, line 29: Docstrings? I don't see any in this class. I think should be added at least for the public functions, since it's important for readability

rre-dataset-generator/src/search_engine/opensearch_engine.py
```diff
@@ -0,0 +1,113 @@
...
+            "query": query,
+            "_source": fields,
+            "size": doc_number
+        }
+
+        return self._search(payload)
+
+    def fetch_for_evaluation(self, query_template: str, doc_fields: List[str], keyword: str = "*") -> List[Document]:
+        query = query_template.replace(self.PLACEHOLDER, keyword)
+        fields = doc_fields if self.UNIQUE_KEY in doc_fields else doc_fields + [self.UNIQUE_KEY]
```
Intrinsical-AI, line 66: What if doc_fields are not in index?

Check: https://github.com/SeaseLtd/rated-ranking-evaluator/pull/189 -> Feel free to adapt or to get inspired

rre-dataset-generator/tests/mocks/opensearch.py
```diff
@@ -0,0 +1,20 @@
+import requests
+
+
+class MockResponseOpenSearchEngine:
+    def __init__(self, hits_data, status_code=200):
```
Intrinsical-AI: typing? What type should be hits_data? Status code accepts floats (200.) for example??

rre-dataset-generator/tests/unit/test_opensearch_engine.py
```diff
@@ -0,0 +1,57 @@
...
+    hit = {
+        "_index": "testcore",
+        "_id": "1",
+        "_score": 1.0,
+        "_source": doc
+    }
+
+    expected_doc = Document(
+        id=doc["id"],
+        fields={k: search_engine._normalize(v) for k, v in doc.items() if k != "id"}
```
Intrinsical-AI: I'd add the _normalize test in a different test file. 

With this tests, if _normalize implementation implements a bug -> gets propagated here -> error consistency -> the test still passing

rre-dataset-generator/tests/unit/test_opensearch_engine.py
```diff
@@ -0,0 +1,57 @@
...
+    )
+
+    def post(*args, **kwargs):
+        return MockResponseOpenSearchEngine([hit], status_code=200)
+
+    monkeypatch.setattr(requests, "post", post)
+
+    result = search_engine.fetch_for_query_generation(
```
Intrinsical-AI: This tests both methods in one single test -> I'd split, since if any of them fails, we don't know which one it's (also no logs)

rre-dataset-generator/tests/mocks/opensearch.py
```diff
@@ -0,0 +1,20 @@
...
+
+    def raise_for_status(self):
+        if self.status_code != 200:
+            raise requests.exceptions.HTTPError(f"Status code: {self.status_code}")
+
+    def json(self):
+        return {
+            "hits": {
+                "total": {"value": len(self._hits_data), "relation": "eq"},
+                "max_score": 1.0,
```
Intrinsical-AI, line 28: I would add strict typing to the return json also - explicit fields and types

rre-dataset-generator/tests/integration/docker-compose.opensearch.yml
```diff
@@ -0,0 +1,33 @@
+services:
+  opensearch:
+    image: opensearchproject/opensearch:3.1.0
+    container_name: opensearch
+    environment:
+      - discovery.type=single-node
+      - node.name=opensearch
+      - bootstrap.memory_lock=true
+      - DISABLE_INSTALL_DEMO_CONFIG=true
+      - DISABLE_SECURITY_PLUGIN=true
```
Intrinsical-AI, line 10: I haven't seem those parmeters before. Why are those config for? Is something custom or needed in general?

rre-dataset-generator/src/search_engine/opensearch_engine.py
```diff
@@ -0,0 +1,113 @@
...
+        Normalize a field value:
+        - string → return [cleaned string]
+        - list of strings → return [cleaned strings]
+        - else → return as it is
+        """
+        if isinstance(value, str):
+            return [clean_text(value)]
+        if isinstance(value, list) and all(isinstance(i, str) for i in value):
+            return [clean_text(i) for i in value]
+        return value
```
Intrinsical-AI: if is not str or list, what happen? No validation? 

 I would add at least a try - except + casting, or a raise ValueError, for example

rre-dataset-generator/tests/unit/test_opensearch_engine.py
```diff
@@ -0,0 +1,57 @@
...
+from src.logger import configure_logging
+from src.model.document import Document
+from src.search_engine.opensearch_engine import OpenSearchEngine
+from tests.mocks.opensearch import MockResponseOpenSearchEngine
+
+configure_logging(level=logging.DEBUG)
+
+
+def test_opensearch_search_engine(monkeypatch):
+    config = Config.load("tests/unit/resources/good_config_opensearch.yaml")
```
Intrinsical-AI: Looks cleaner to me with a config fixture common for all methods. 
EJ:


```
import pytest
from src.config import Config
@pytest.fixture
def opensearch_config():
    return Config.load("tests/unit/resources/good_config_opensearch.yaml")
```
And then, in each method:
```
def test_opensearch_search_engine(monkeypatch, opensearch_config):
```


Bonus: testing different config files with parametrization

```
@pytest.fixture(params=[
    "tests/unit/resources/good_config_opensearch.yaml",
    "tests/unit/resources/edge_case_config.yaml"
])
def opensearch_config(request):
    return Config.load(request.param)
```

rre-dataset-generator/tests/unit/resources/good_config_opensearch.yaml
Intrinsical-AI, line 1: this is common for all of the implementations, but I just realized that this maybe could lead to string injection or even code exection. 

E.g: processing a dataset of OpenSearch / Elastic / Vespa Queries

The string should be parsed and filtered of - operators (AND / OR / WHERE, script tags.. and that stuff. We should agree on how to do it for the diferent engines. TBD

rre-dataset-generator/tests/mocks/opensearch.py
```diff
@@ -0,0 +1,20 @@
+import requests
+
+
+class MockResponseOpenSearchEngine:
+    def __init__(self, hits_data, status_code=200):
+        self._hits_data = hits_data if isinstance(hits_data, list) else [hits_data]
```
Intrinsical-AI: Without strict typing I don't know if this is robust

Intrinsical-AI (commented):
Request to add docstrings, typing and validations. Suggestion of adding pytests fixtures, and spliting the tests individually

rre-dataset-generator/src/search_engine/opensearch_engine.py
nseidan: see `BaseSearchEngine` docstring, this class inherits the docstring from the base.

rre-dataset-generator/src/search_engine/opensearch_engine.py
nseidan: opensearch handles this, if the given fields are not present, still opensearch returns docs with ids and other existing fields under `_source`

rre-dataset-generator/tests/integration/docker-compose.opensearch.yml
nseidan: common params to disable security checks in dev and not to include demo config at node/cluster startup, which includes admin/hardcoded password with roles etc.

rre-dataset-generator/src/search_engine/opensearch_engine.py
nseidan: updated

rre-dataset-generator/tests/mocks/opensearch.py
nseidan: updated

rre-dataset-generator/tests/mocks/opensearch.py
nseidan: updated

rre-dataset-generator/tests/mocks/opensearch.py
nseidan: added

rre-dataset-generator/tests/unit/test_opensearch_engine.py
nseidan: updated

rre-dataset-generator/tests/unit/test_opensearch_engine.py
nseidan: added

rre-dataset-generator/tests/unit/test_opensearch_engine.py
nseidan: adjusted

rre-dataset-generator/src/search_engine/opensearch_engine.py
```diff
@@ -0,0 +1,126 @@
...
+        except (ConnectionError, Timeout, RequestException, HTTPError) as e:
+            log.error(f"OpenSearch query failed: {e}\nPayload: {payload}")
+            raise
+
+        hits = response.json().get("hits", {}).get("hits", [])
+        result = []
+
+        for hit in hits:
+            source = hit.get("_source", {})
+            doc_id = source.get("id", hit.get("_id"))
```
nicolo-rinaldi, line 95: "_id" is `self.UNIQUE_KEY` right?

rre-dataset-generator/tests/integration/docker-compose.opensearch.yml
Intrinsical-AI: Ok, told the same to Nicolo. When we adapt to real envs (if so), will have to remember this as could be important

rre-dataset-generator/tests/mocks/opensearch.py
Intrinsical-AI: :+1:

rre-dataset-generator/tests/unit/resources/good_config_opensearch.yaml
Intrinsical-AI: (Not blocking from my part since Nicolo it's similar and for mocks / tests should mostly work)

rre-dataset-generator/src/search_engine/opensearch_engine.py
Intrinsical-AI: Ok thats good to know. I'd add it as comment or something, as it's very particular (and useful!).

 Anw: I would add then need a funtion to print/log/process/filter the fields based on the "_source" string -> So we get infromation of if the fields searched by the user are present. Minmal simple log would be useful

rre-dataset-generator/src/search_engine/opensearch_engine.py
Intrinsical-AI: That doesn't make fully sense to me, as the base docstring describes the contract, but not how it's implemented. I don't know if it's a convetion or smth? Google Style for Docstring maybe?

nseidan: for me it makes sense, since docstring is more of description what the method is doing , not about implementation how it is done.

rre-dataset-generator/tests/integration/docker-compose.opensearch.yml
nseidan: no need to worry about this; in real cases, search engine comes from users.  This test env is just for us to stimulate.

rre-dataset-generator/src/search_engine/opensearch_engine.py
nseidan: _id -> auto generated and unique key

rre-dataset-generator/src/search_engine/opensearch_engine.py
nseidan: in the search method is already done.

nicolo-rinaldi (dismissed):

rre-dataset-generator/src/search_engine/opensearch_engine.py
```diff
@@ -0,0 +1,126 @@
...
+
+class OpenSearchEngine(BaseSearchEngine):
+    """
+    OpenSearch implementation to search in a given index.
+    """
+
+    def __init__(self, endpoint: HttpUrl):
+        super().__init__(endpoint)
+        self.HEADERS = {'Content-Type': 'application/json'}
+        self.UNIQUE_KEY = "id"
```
dantuzi, line 24: I recall that the unique key for OpenSearch was `_id`, isn't it?

rre-dataset-generator/src/search_engine/opensearch_engine.py
Intrinsical-AI: This doesn't make any sense to me, but goes against documentation principles given this is not a trivial implementation, but the core logic of the engine to be honest.. maybe you can share some references / sources?

rre-dataset-generator/src/search_engine/opensearch_engine.py
Intrinsical-AI: No, it's not. I mean log specific information on what fields are missing. From what I can see it's not implemented.

rre-dataset-generator/tests/unit/test_opensearch_engine.py
Intrinsical-AI: Good! :+1:

rre-dataset-generator/src/search_engine/opensearch_engine.py
nseidan: "id" is in the index, so `_id` and `id` are the same, so both are unique keys.

dantuzi (dismissed):

rre-dataset-generator/src/search_engine/opensearch_engine.py
nseidan: updated

rre-dataset-generator/src/search_engine/opensearch_engine.py
nseidan: added minimal docstrings

Intrinsical-AI (approved):

nicolo-rinaldi (approved):

dantuzi (approved):

PR #185, opened 2025-07-25
DAGE-20: added ElasticSearch support

nicolo-rinaldi:
- updated SearchEngineFactory with elasticsearch support
- created elasticsearch_search_engine.py file with the class needed for the integration and related unit test file
- changed the names of the files used for test in the resources folder

[Jira ticket](https://sease.atlassian.net/browse/DAGE-20)


rre-dataset-generator/tests/integration/elasticsearch-init/Dockerfile
```diff
@@ -0,0 +1,4 @@
+FROM docker.elastic.co/elasticsearch/elasticsearch:9.0.4
```
nicolo-rinaldi: @nseidan why this image ? it's the full search engine image , better to have a lightweight one

nicolo-rinaldi: What do you propose? In the meanwhile, I search a little bit for some alternatives

rre-dataset-generator/tests/integration/elasticsearch-init/Dockerfile
```diff
@@ -0,0 +1,4 @@
+FROM curlimages/curl:8.14.1
```
nicolo-rinaldi, line 1: This is the image @nseidan used. Seems to be very light since we need to do only `curl` operations

rre-dataset-generator/tests/integration/elasticsearch-init/Dockerfile
nseidan: in solr as well as in opensearch the curl image is used as we used curl inside docker in `solr-init.sh`/`opensearch-init.sh`

rre-dataset-generator/tests/integration/elasticsearch-init/Dockerfile
nseidan: it's about 9 MB in size

rre-dataset-generator/src/search_engine/elasticsearch_search_engine.py
```diff
@@ -0,0 +1,120 @@
...
+        except RequestException as e:
+            log.error(f"Unexpected error during request to {search_url}\nError: {e}")
+            raise RequestException(f"Unexpected error during request to {search_url}\nError: {e}")
+
+        match response.status_code:
+            case 200:
+                log.debug("Elasticsearch query successful.")
+                log.debug(f"URL: {search_url}")
+                log.debug(f"Payload: {payload}")
+                # log.debug(f"Response: {response.json()}")
```
nseidan: is this needed  # log.debug(f"Response: {response.json()}") ?

nseidan (dismissed):

rre-dataset-generator/src/search_engine/elasticsearch_search_engine.py
nicolo-rinaldi: removed

rre-dataset-generator/src/search_engine/elasticsearch_search_engine.py
```diff
@@ -0,0 +1,119 @@
...
+            }
+        payload["_source"] = doc_fields
+        return self._search(payload)
+
+    def _search(self, payload: Dict[str, Any]) -> List[Document]:
+        """Search for documents using a query."""
+        search_url = urljoin(self.endpoint.encoded_string(), '_search')
+
+        try:
+            response = requests.post(search_url, headers=self.HEADERS, json=payload)
```
Intrinsical-AI, line 114: allow_redirects=False (timeout also nice to avoid hanging)

rre-dataset-generator/src/search_engine/elasticsearch_search_engine.py
```diff
@@ -0,0 +1,119 @@
...
+        except RequestException as e:
+            log.error(f"Unexpected error during request to {search_url}\nError: {e}")
+            raise RequestException(f"Unexpected error during request to {search_url}\nError: {e}")
+
+        match response.status_code:
+            case 200:
+                log.debug("Elasticsearch query successful.")
+                log.debug(f"URL: {search_url}")
+                log.debug(f"Payload: {payload}")
+                raw_docs = response.json()['hits']['hits']
```
Intrinsical-AI:
```
data = response.json()
raw_docs = (data.get('hits') or {}).get('hits') or [] 

```
-> Should be safer to unexpected responses

.get in general it's more safe / robust, as should not rise errors, while direct access ['hits'] -> raises error if not present

rre-dataset-generator/src/search_engine/elasticsearch_search_engine.py
```diff
@@ -0,0 +1,119 @@
...
+        if keyword:
+             payload = json.loads(query_template.replace(self.PLACEHOLDER, keyword))
+        else:
+            payload = {
+            "query": {"match_all": {}}
+            }
+        payload["_source"] = doc_fields
+        return self._search(payload)
+
+    def _search(self, payload: Dict[str, Any]) -> List[Document]:
```
Intrinsical-AI, line 101: This is completly different from my implementation, and from Naz. Maybe we should align the search engine logic

rre-dataset-generator/tests/mocks/elasticsearch.py
```diff
@@ -0,0 +1,11 @@
+class MockResponseElasticsearchEngine:
+    def __init__(self, json_data, status_code=200):
```
Intrinsical-AI: Typing?

rre-dataset-generator/tests/integration/docker-compose-elasticsearch.yml
```diff
@@ -0,0 +1,16 @@
+services:
+  elasticsearch:
+    image: docker.elastic.co/elasticsearch/elasticsearch:9.0.4
+    container_name: elasticsearch_test
+    ports:
+      - "9200:9200"
+    environment:
+      - discovery.type=single-node
+      - xpack.security.enabled=false
+      - xpack.security.enrollment.enabled=false
```
Intrinsical-AI: Why is that for? What is it turning off?

rre-dataset-generator/src/search_engine/elasticsearch_search_engine.py
```diff
@@ -0,0 +1,119 @@
...
+                log.debug(f"URL: {search_url}")
+                log.debug(f"Payload: {payload}")
+                raw_docs = response.json()['hits']['hits']
+                reformat_raw_doc = []
+                for doc in raw_docs:
+                    clean_doc = dict()
+                    clean_doc['id'] = doc[self.UNIQUE_KEY]
+                    clean_doc['fields'] = doc['_source']
+                    reformat_raw_doc.append(Document(**clean_doc))
+                return reformat_raw_doc
```
Intrinsical-AI: Tbh: if we agree on standarizing the logic between engines, I'd follow the [PR](https://github.com/SeaseLtd/rated-ranking-evaluator/compare/dataset-generator...DAGE-19_opensearch_support) style, as seems cleaner. 

That said, this approach enhances capturing exceptions in detail, which I find useful. Maybe we could do in a cleaner way. TBD

rre-dataset-generator/tests/mocks/elasticsearch.py
```diff
@@ -0,0 +1,11 @@
...
+        self._json_data = json_data
+        self.status_code = status_code
+
+    def json(self):
+        return {
+            "hits": {
+                "hits": self._json_data if isinstance(self._json_data, list) else [self._json_data]
```
Intrinsical-AI: The point it's that without strict typing I don't know if this is correct. Maybe not critical

rre-dataset-generator/tests/unit/test_elasticsearch_search_engine.py
```diff
@@ -0,0 +1,77 @@
...
+
+    # apply the monkeypatch for requests.post to mock_post
+    monkeypatch.setattr(requests, "post",
+                        lambda *args, **kwargs: MockResponseElasticsearchEngine(mock_doc, status_code=200)
+                        )
+    # search_engine.extract_documents_to_generate_queries, which contains requests.post, uses the monkeypatch
+    result = search_engine.fetch_for_query_generation(documents_filter=config.documents_filter,
+                                                      doc_number=config.doc_number,
+                                                      doc_fields=config.doc_fields)
+    assert result[0] == Document(**mock_dict)
```
Intrinsical-AI, line 55: Split the query_generation ad the scoring in 2 different tests makes sense for me

rre-dataset-generator/tests/unit/test_elasticsearch_search_engine.py
```diff
@@ -0,0 +1,77 @@
...
+
+from src.search_engine.elasticsearch_search_engine import ElasticsearchSearchEngine
+from src.model.document import Document
+import logging
+
+configure_logging(level=logging.DEBUG)
+
+
+def test_elasticsearch_search_engine(monkeypatch):
+    config = Config.load("tests/unit/resources/elasticsearch_good_config.yaml")
```
Intrinsical-AI: fixture common for the tests?

rre-dataset-generator/src/search_engine/elasticsearch_search_engine.py
```diff
@@ -0,0 +1,119 @@
...
+            "query": query,
+            "_source": doc_fields
+        }
+
+        return self._search(payload)
+
+    def fetch_for_evaluation(self, query_template: str, doc_fields: List[str], keyword: str=None) -> List[Document]:
+        """Search for documents using a query."""
+        if keyword:
+             payload = json.loads(query_template.replace(self.PLACEHOLDER, keyword))
```
Intrinsical-AI: this is common for all of the implementations, but I just realized that this could allow string injection or even - in the worst case - code exection (the dataset-generator projet might work with code snippets as a database), as it's getting passed directly to the engine 

```
payload = json.loads(query_template.replace(self.PLACEHOLDER, keyword)) 
> query_template.replace(self.PLACEHOLDER, keyword)
```
 If keyword contains "o \n, you break the string (because we are replacing text, not a JSON value). 


I'm adding some parsing / cleaning for the Vespa. Let's discuss

rre-dataset-generator/src/search_engine/elasticsearch_search_engine.py
```diff
@@ -0,0 +1,119 @@
...
+        super().__init__(endpoint)
+        self.HEADERS = {'Content-Type': 'application/json'}
+        log.debug(f"Working on endpoint: {self.endpoint}")
+        self.UNIQUE_KEY = "_id"
+
+    def fetch_for_query_generation(self,
+                                   documents_filter: Union[None, List[Dict[str, List[str]]]],
+                                   doc_number: int,
+                                   doc_fields: List[str]) -> List[Document]:
+
```
Intrinsical-AI, line 40: I'd add docstrings at least for the public functions

rre-dataset-generator/tests/unit/test_solr_search_engine.py
```diff
@@ -47,10 +47,10 @@ def test_solr_search_engine(monkeypatch):
     assert result[0] == Document(**mock_dict)
 
 def test_solr_search_engine_negative_post(monkeypatch):
-    config = Config.load("tests/unit/resources/good_config.yaml")
+    config = Config.load("tests/unit/resources/solr_good_config.yaml")
```
Intrinsical-AI: Looks cleaner to me with a config fixture common for all methods.
EJ:
```

import pytest
from src.config import Config
@pytest.fixture
def opensearch_config():
    return Config.load("tests/unit/resources/good_config_opensearch.yaml")
```
And then, in each method:

```
def test_opensearch_search_engine(monkeypatch, opensearch_config):
    ....
    
```

Intrinsical-AI: Bonus: testing different config files with parametrization
```

@pytest.fixture(params=[
    "tests/unit/resources/good_config_opensearch.yaml",
    "tests/unit/resources/edge_case_config.yaml"
])
def opensearch_config(request):
    return Config.load(request.param)
```

rre-dataset-generator/tests/unit/test_elasticsearch_search_engine.py
```diff
@@ -0,0 +1,77 @@
...
+        search_engine = ElasticsearchSearchEngine("https://fakeurl")
+
+        with pytest.raises(HTTPError):
+            search_engine.fetch_for_query_generation(
+                documents_filter=config.documents_filter,
+                doc_number=config.doc_number,
+                doc_fields=config.doc_fields
+            )
+
+        with pytest.raises(HTTPError):
```
Intrinsical-AI, line 95: I'd split in 2 testing methods so we know which one it's failing

rre-dataset-generator/src/search_engine/elasticsearch_search_engine.py
```diff
@@ -0,0 +1,119 @@
...
+        filter_clauses = []
+        if documents_filter is not None:
+            for dict_field in documents_filter:
+                for field, values in dict_field.items():
+                    if not values:
+                        continue
+                    if len(values) == 1:
+                        filter_clauses.append({"term": {field: values[0]}})
+                    else:
+                        filter_clauses.append({"terms": {field: values}})
```
Intrinsical-AI: If values it's a list of strings (without checking / validation), I'd try to assure that before passing to this function, it's alredy checked / validated -> If not this could lead  to errors

Intrinsical-AI (commented):
I'd add parsing, typing, docstrings, and use pytests fixtures. 
 
 Also: I think it's needed to follow a general common implementation for the engines, as we're doing very diferent each one (further than syntax specifics)

rre-dataset-generator/src/search_engine/elasticsearch_search_engine.py
Intrinsical-AI: A more robust solution might be:

1. First we load the raw template
2. Then we fill the placeholder field as VALUE (not handled as text replacement)

```
 tmpl = json.loads(query_template)

if (
        isinstance(tmpl, dict) and
        isinstance(tmpl.get("query"), dict) and
        isinstance(tmpl["query"].get("match"), dict) and
        tmpl["query"]["match"].get("message") == PLACEHOLDER
    ):
        tmpl["query"]["match"]["message"] = keyword
    else:
        raise ValueError("The placeholder is not where expected")
```

rre-dataset-generator/src/search_engine/elasticsearch_search_engine.py
nicolo-rinaldi: This was exactly how I did in solr btw

nicolo-rinaldi: Not going to change it until we decide some common ground, doesn't make sense to change it more than once

rre-dataset-generator/src/search_engine/elasticsearch_search_engine.py
nicolo-rinaldi: changed

rre-dataset-generator/src/search_engine/elasticsearch_search_engine.py
nicolo-rinaldi: added

rre-dataset-generator/tests/integration/docker-compose-elasticsearch.yml
nicolo-rinaldi: There are security settings for a production environment in elasticsearch. For a test env, since we do not need to deal with that, it is much easier to just turn them off.

rre-dataset-generator/tests/mocks/elasticsearch.py
nicolo-rinaldi: added

rre-dataset-generator/tests/unit/test_solr_search_engine.py
nicolo-rinaldi: changed with `@pytest.fixture`

rre-dataset-generator/tests/unit/test_elasticsearch_search_engine.py
nicolo-rinaldi: done for tests that raises an error

rre-dataset-generator/src/search_engine/elasticsearch_search_engine.py
nicolo-rinaldi: Let's bring up this tomorrow at the standup

rre-dataset-generator/tests/unit/test_elasticsearch_search_engine.py
nicolo-rinaldi: done

rre-dataset-generator/tests/mocks/elasticsearch.py
nicolo-rinaldi: added typing and updated tests accordingly

rre-dataset-generator/tests/unit/test_elasticsearch_search_engine.py
Intrinsical-AI: But, if one raise the error and the other doesn't, which one is which?

rre-dataset-generator/src/search_engine/elasticsearch_search_engine.py
Intrinsical-AI: Added. I resolve it

rre-dataset-generator/src/search_engine/elasticsearch_search_engine.py
Intrinsical-AI: I'm not saying you did wrong, I'm saying as Naz and you based on the feedback, inteded to align in structure, makes sense to also align in structure

rre-dataset-generator/tests/integration/docker-compose-elasticsearch.yml
Intrinsical-AI: Oka, the point it's that I didn't need to turn off anything for Vespa, but seems to be working. Also noticed that Naz also disabled some security configs in her dockerfile

 When / if we have to test integrations for prod we might need a different setup. Assumed it's not needed, but rather to notice now

rre-dataset-generator/src/search_engine/elasticsearch_search_engine.py
nicolo-rinaldi: nvm, I updated. I think your way is way more clear. Changed for both solr and elastic

rre-dataset-generator/tests/unit/test_elasticsearch_search_engine.py
nicolo-rinaldi: i splitted

rre-dataset-generator/src/search_engine/elasticsearch_search_engine.py
Intrinsical-AI: Mine and Naz are not 100% aligned either, but anyway I really appreciate the effort. You're a beast

 We could make a meeting tomorrow to align refine + polish, let's discuss in the standup or afterwards :+1:

rre-dataset-generator/tests/integration/docker-compose-elasticsearch.yml
nicolo-rinaldi: I've seen this done a lot of times, so I replicated it

rre-dataset-generator/tests/unit/test_elasticsearch_search_engine.py
nicolo-rinaldi: done

rre-dataset-generator/src/search_engine/elasticsearch_search_engine.py
nicolo-rinaldi: uniformed for both solr and elastic

rre-dataset-generator/src/search_engine/solr_search_engine.py
```diff
@@ -80,66 +92,75 @@ def fetch_for_query_generation(self,
...
-            case 404:
-                error_msg = f"404 Not Found: The Solr endpoint was not found.\nURL: {search_url}"
-            case 500:
-                error_msg = f"500 Internal Server Error: Solr encountered a problem.\nURL: {search_url}"
-            case _:
-                error_msg = f"Unexpected status code {response.status_code}.\nURL: {search_url}\nPayload: {payload}"
-        log.error(error_msg)
-        raise HTTPError(error_msg)
+            response.raise_for_status()
+        except (ConnectionError, Timeout, RequestException, HTTPError) as e:
```
Intrinsical-AI, line 127: Nice catch :+1:

rre-dataset-generator/src/search_engine/solr_search_engine.py
```diff
@@ -80,66 +92,75 @@ def fetch_for_query_generation(self,
...
+                 key: self._normalize(value)
+                 for key, value in hit.items()
+                 if key != self.UNIQUE_KEY
+             }
+
+             result.append(Document(id=doc_id, fields=fields))
+        return result
+
+    @staticmethod
+    def _normalize(value: Any) -> List[str]:
```
Intrinsical-AI, line 145: Clean, like the readability

rre-dataset-generator/tests/mocks/solr.py
```diff
@@ -1,15 +1,24 @@
+from typing import List
+import requests
+
 class MockResponseSolrEngine:
-    def __init__(self, json_data, status_code=200):
+    def __init__(self, json_data: List, status_code: int =200):
```
Intrinsical-AI, line 5: I like this! :+1:

rre-dataset-generator/tests/unit/test_solr_search_engine.py
```diff
@@ -14,58 +14,84 @@
 
 configure_logging(level=logging.DEBUG)
 
+@pytest.fixture
+def solr_config():
```
Intrinsical-AI, line 18: I think this is way better to extend more tests afterwards if needed :+1:

rre-dataset-generator/src/search_engine/elasticsearch_search_engine.py
Intrinsical-AI: I consider this discussed an agreed, so I close the conversation

rre-dataset-generator/src/search_engine/elasticsearch_search_engine.py
Intrinsical-AI: This is not bocking, only a remember for the future

Intrinsical-AI (dismissed):
I think this is a very good refactor!   I'm glad you decided to add it, thanks for the openess! :100:

Intrinsical-AI (dismissed):

rre-dataset-generator/src/search_engine/elasticsearch_search_engine.py
```diff
@@ -0,0 +1,160 @@
...
+        query = {"match_all": {}}
+
+        # Add filters, if provided
+        filter_clauses = []
+        if documents_filter is not None:
+            for dict_field in documents_filter:
+                for field, values in dict_field.items():
+                    if not values:
+                        continue
+                    if len(values) == 1:
```
dantuzi: Maybe it is more readable if we avoid this check and we always pass `filter_clauses.append({"terms": {field: values}})`

rre-dataset-generator/src/search_engine/solr_search_engine.py
dantuzi: I like it too but I'd suggest, for the next time, to use a different PR to not mixup too many things

rre-dataset-generator/tests/integration/docker-compose.solr.yml
dantuzi, line 1: I like it but, for the next time, please do refactoring and fixes in a separated Pr

rre-dataset-generator/tests/unit/test_elasticsearch_search_engine.py
```diff
@@ -0,0 +1,104 @@
...
+    }
+
+@pytest.fixture
+def mock_dict(mock_doc):
+    return {
+        'id': mock_doc['_id'],
+        'fields': mock_doc["_source"]
+    }
+
+def test_elasticsearch_search_engine_fetch_for_query_generation(monkeypatch, elasticsearch_config, mock_doc, mock_dict):
```
dantuzi: Can we use the same test name format we are using everywhere else?
def `test_<what I'm testing>__expect__<what I'm expecting>`

rre-dataset-generator/tests/unit/test_elasticsearch_search_engine.py
```diff
@@ -0,0 +1,104 @@
...
+        search_engine = ElasticsearchSearchEngine("https://fakeurl")
+
+        with pytest.raises(HTTPError):
+            search_engine.fetch_for_evaluation(
+                keyword="and",
+                query_template=elasticsearch_config.query_template,
+                doc_fields=elasticsearch_config.doc_fields
+            )
+
+def test_elasticsearch_search_engine_bad_url_EXPECTED_ValidationError():
```
dantuzi: Initially we were using the word "EXPECT" (uppercased) as a separator but the upper case words are not standard in python so we are moving to "__expect__" (with the double underscore before and after the word expect)

rre-dataset-generator/src/search_engine/elasticsearch_search_engine.py
nicolo-rinaldi: changed

rre-dataset-generator/tests/unit/test_elasticsearch_search_engine.py
nicolo-rinaldi: done

rre-dataset-generator/tests/unit/test_elasticsearch_search_engine.py
nicolo-rinaldi: done

dantuzi (approved):

PR #189, opened 2025-07-28
Added log for avoid silent incoherence between user input and app behaviour - non-existing fields

Intrinsical-AI:
Added:

```
        # Solr default behavior, passing nonexistent fields results in a silent failure with no logging -- added logging
        self.schema_fields = {field['name'] for field in requests.get(urljoin(self.endpoint.encoded_string(), 'schema/fields')).json()['fields']}
        log.debug(f"Schema fields loaded: {len(self.schema_fields)} fields found.")
```

This: 1)notice in the logs which fields are present in the index. 2) Subset only the valid ones. Might need aditional logic to handle empty cases. let me know if you agree with the approach.

### DIFF
--- a/rre-dataset-generator/src/search_engine/solr_search_engine.py
+++ b/rre-dataset-generator/src/search_engine/solr_search_engine.py
@@ -23,6 +23,9 @@ class SolrSearchEngine(BaseSearchEngine):
         log.debug(f"Working on endpoint: {self.endpoint}")
         self.UNIQUE_KEY = requests.get(urljoin(self.endpoint.encoded_string(), 'schema/uniquekey')).json()['uniqueKey']
         log.debug(f"uniqueKey found: {self.UNIQUE_KEY}")
+        # Solr default behavior, passing nonexistent fields results in a silent failure with no logging -- added logging
+        self.schema_fields = {field['name'] for field in requests.get(urljoin(self.endpoint.encoded_string(), 'schema/fields')).json()['fields']}
+        log.debug(f"Schema fields loaded: {len(self.schema_fields)} fields found.")
 
     def _template_to_json_payload(self, template_payload: str) -> Dict[str, Any]:
         """
--- a/rre-dataset-generator/src/search_engine/solr_search_engine.py
+++ b/rre-dataset-generator/src/search_engine/solr_search_engine.py
@@ -18,6 +18,12 @@ class SolrSearchEngine(BaseSearchEngine):
     Solr implementation to search into a given collection
     """
     def __init__(self, endpoint: HttpUrl):
+        """ 
+        ENDPOINTS:
+            http://localhost:8983/solr/testcore/schema/uniquekey
+            http://localhost:8983/solr/testcore/schema/fields
+            http://localhost:8983/solr/testcore/schema/dynamicfields
+        """
         super().__init__(endpoint)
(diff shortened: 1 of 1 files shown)


Intrinsical-AI:
Test are not passing now since we have filtered the fields.. Incoming commit

Intrinsical-AI:
Now test are passing. We stored a set of strings as a class attribute (`self.schema_fields`, that requests the Solr index (given by the endpoint parameter), the present fields. 

Included debug logs following the constructor pattern.

nicolo-rinaldi (approved):

Intrinsical-AI:
@dantuzi Lmk what you think and I'll rebase if needed

rre-dataset-generator/src/search_engine/solr_search_engine.py
```diff
@@ -23,6 +23,17 @@ def __init__(self, endpoint: HttpUrl):
         log.debug(f"Working on endpoint: {self.endpoint}")
         self.UNIQUE_KEY = requests.get(urljoin(self.endpoint.encoded_string(), 'schema/uniquekey')).json()['uniqueKey']
         log.debug(f"uniqueKey found: {self.UNIQUE_KEY}")
+        # Solr default behavior, passing nonexistent fields results in a silent failure with no logging -- added logging
+        try:
+            # Ask the endpoint the existing fields
+            schema_resp = requests.get(urljoin(self.endpoint.encoded_string(), 'schema/fields')).json()
+            # subset only the present fields
+            self.schema_fields = {field['name'] for field in schema_resp.get('fields', [])}
```
dantuzi: I don't see `self.schema_fields` used somewhere. Am I wrong?

rre-dataset-generator/src/search_engine/solr_search_engine.py
```diff
@@ -23,6 +23,17 @@ def __init__(self, endpoint: HttpUrl):
...
+        except Exception as exc:
+            # In unit tests the mock response may not contain the expected structure -> store empty set
+            log.debug(f"Schema fields endpoint did not return expected payload: {exc}; defaulting to empty schema list")
+            self.schema_fields = set()
+        log.debug(f"Schema fields loaded: {len(self.schema_fields)} fields found.")
```
dantuzi, line 46: It looks like this entire PR is for writing this string in the logs. Why do you think this is useful?

rre-dataset-generator/src/search_engine/solr_search_engine.py
dantuzi: If this is needed only in this small piece of code, `schema_fields` should be a local variable

Intrinsical-AI: I think it might be useful for checking the payload when building the queries to the engine

rre-dataset-generator/src/search_engine/solr_search_engine.py
Intrinsical-AI: I as developer / user rather knowing when some fields are present or not, than just executing and expect to behave nice. The problem was the silent omission of fields if not present on the schema

Intrinsical-AI: Another option could be implementing this directly on the _search method. Maybe makes more sense. But I'd like to know when the searched fields are not present

PR #193, opened 2025-08-01
Dage 37: Datastore re-allocated, Query Context with Pydantic

Intrinsical-AI:
### **Task**: [DAGE-37](https://sease.atlassian.net/browse/DAGE-37?atlOrigin=eyJpIjoiMTNkOTNjNWRmZTEzNDNmMjg1YTgxMjBkNTcwYjM5NDEiLCJwIjoiaiJ9)

(Check the task description for a summary before checking the files)

### **refactor: datastore → Pydantic models, relocated module, and API cleanup**

This PR simplifies and strengthens the `DataStore` layer by:

* Replacing the custom `QueryRatingContext` with validated Pydantic models (`Rating`, `Query`, `Document`).
* Relocating `data_store.py` from `search_engine/` to top-level `src/` for a stable import path.
* Updating writers and core modules (`dataset_generator.py`, writers,..) to use the new API.
* Cleaning up legacy code and aligning tests with the new structure.


rre-dataset-generator/src/data_store.py
```diff
@@ -0,0 +1,134 @@
...
+        for rating_id in self.queries[query_id].related_ratings_ids:
+            rating = self.ratings[rating_id]
+            if rating.doc_id == doc_id:
+                return rating.score
+        return None
+    
+    # ---------- commands ----------
+    def add_doc(self, doc: Document) -> None:
+        if doc.id in self.docs:
+            log.error(f"Document {doc.id} already exists")
```
nicolo-rinaldi: **[curiosity]**

Do we want to log an error or a warning?

rre-dataset-generator/src/model/__init__.py
nicolo-rinaldi, line 1: Why do we want to have the ratings also in the query? The Rating is storing them. Maybe we can avoid having duplicates and when we want all the ratings from a query we look into Rating. I don't know, maybe I'm not seeing the whole image.

nicolo-rinaldi (dismissed):

rre-dataset-generator/src/data_store.py
Intrinsical-AI: I choose error on purpose. Why should we try adding already indexed docs? It's better to prefilter IMO

rre-dataset-generator/src/writers/abstract_writer.py
```diff
@@ -1,34 +1,16 @@
...
-        """
-        result = []
-        for query_ctx in self.datastore.get_queries():
-            query_text = query_ctx.get_query_text()
-            for doc_id in query_ctx.get_doc_ids():
-                if query_ctx.has_rating_score(doc_id):
-                    rating = query_ctx.get_rating_score(doc_id)
-                    result.append((query_text, doc_id, rating))
-        return result
+        pass
```
nicolo-rinaldi, line 16: I like this

rre-dataset-generator/dataset_generator.py
```diff
@@ -34,53 +24,47 @@ def get_and_setup_logging(verbose: bool = False) -> Logger:
...
         query_response: LLMQueryResponse = llm_service.generate_queries(doc, num_queries_per_doc)
         for query_text in query_response.get_queries():
             if len(data_store.get_queries()) >= config.num_queries_needed:
                 return
-            query_id: str = data_store.add_query(query_text, doc.id)
-            data_store.add_rating_score(query_id, doc.id, max(config.relevance_label_set),
-                                        "Default max rating is assigned because the query is generated by the document")
+            query = Query(text=query_text)
+            query_id: str = data_store.add_query(query)
+            data_store.create_rating_score(query_id, doc.id, max(config.relevance_label_set))
```
nicolo-rinaldi: why here you removed the "hard-coded" explanation?

rre-dataset-generator/dataset_generator.py
```diff
@@ -95,27 +79,22 @@ def add_cartesian_product_scores(config: Config, data_store: DataStore, llm_serv
...
-    add_cartesian_product_scores(config, data_store, service)
+    retrieve_and_add_documents(config, data_store)
 
-    writer.write(config.output_destination)
+    add_cartesian_product_scores(service, config, data_store)
 
-    log.info(f"Synthetic Dataset has been generated in: {config.output_destination}")
+    writer.write(config.output_destination, data_store)
 
-    if config.save_llm_explanation:
```
nicolo-rinaldi, line 125: you are no more saving the file with explanations

nicolo-rinaldi (commented):
After we solve the explanation stuff, I'll approve.

rre-dataset-generator/dataset_generator.py
Intrinsical-AI: You're right, added the method to the new datastore but not invoking it. Will update

rre-dataset-generator/src/model/__init__.py
Intrinsical-AI: Should be resolved in last version

rre-dataset-generator/src/data_store.py
```diff
@@ -0,0 +1,274 @@
...
+    """
+
+    def __init__(self, path: Path = TMP_FILE, ignore_saved_data: bool = False):
+        self.path = path
+
+        # Primary (id → object)
+        self.docs: Dict[str, Document] = {}
+        self.queries: Dict[str, Query] = {}
+
+        # Simplified ratings storage for MVP
```
dantuzi: MVP?

rre-dataset-generator/src/data_store.py
```diff
@@ -0,0 +1,274 @@
...
+
+    def get_ratings(self) -> List[Rating]:
+        """Gets all ratings. Complexity: O(P) where P is the number of ratings."""
+        return list(self.rating_by_pair.values())
+
+    def get_ratings_for_query(self, query_id: str) -> List[Rating]:
+        """(MVP) Returns ratings for a query. Complexity: O(P) where P is the number of ratings."""
+        if not self.has_query(query_id):
+            return []
+        return [r for r in self.rating_by_pair.values() if r.query_id == query_id]
```
dantuzi: Have you thought of defining `self.rating_by_pair = dict[query_id, dict[doc_id, Rating]]`? 
In this specific case, you can quickly extract the list of ratings for a single query by calling `self.rating_by_pair.get(query_id).values()`

rre-dataset-generator/src/data_store.py
Intrinsical-AI: ?

rre-dataset-generator/src/data_store.py
Intrinsical-AI: Hey Daniele, thanks for the suggestion!

Before changing the data structure here, could you share the concrete wins you have in mind (and why it’s needed in this PR)?

For example: which path is a bottleneck today, the expected speedup at our current k/P, the methods/call-sites that would change (has_rating_score, get_rating_score, load/save, clear, …)?

I rather keep this PR ""minimal"", but if you specify the constraints, I'm sure we can improve it :+1:

nicolo-rinaldi (approved):

rre-dataset-generator/dataset_generator.py
```diff
@@ -34,58 +24,50 @@ def get_and_setup_logging(verbose: bool = False) -> Logger:
...
-
 def add_user_queries(config: Config, data_store: DataStore):
     if config.queries is not None:
         with open(config.queries, 'r', encoding='utf-8') as file:
             for line in file:
-                cleaned = line.strip()
-                if cleaned:
-                    data_store.add_query(cleaned)
-
+                if line.strip():
```
dantuzi: With the current solution, the method `strip` is called twice. As a general rule, it's better to save the stripped string in a variable instead of calling the same functionality twice

rre-dataset-generator/.gitignore
```diff
@@ -20,3 +20,4 @@ output/
 # Environment variables
 .env
 
+**/tmp
```
dantuzi, line 23: I don't like the .gitignore linking a folder in the parent folder.
Isn't it better to create the temporary folder directly in the `rre-dataset-generator`?

rre-dataset-generator/src/config.py
```diff
@@ -43,18 +44,16 @@ def check_no_empty_fields(cls, v):
 
     @field_validator('queries')
     def check_doc_type(cls, v):
```
dantuzi, line 47: I missed the variable name `v` before. Does `v` stand for `value`? Do you mind replacing it with a more meaningful name? Thank you

rre-dataset-generator/src/data_store.py
```diff
@@ -0,0 +1,274 @@
...
+    def __init__(self, path: Path = TMP_FILE, ignore_saved_data: bool = False):
+        self.path = path
+
+        # Primary (id → object)
+        self.docs: Dict[str, Document] = {}
+        self.queries: Dict[str, Query] = {}
+
+        # Simplified ratings storage
+        self.rating_by_pair: Dict[Tuple[str, str], Rating] = {}    # (query_id, doc_id) → Rating
+        self.docs_by_query: Dict[str, Set[str]] = defaultdict(set) # query_id → doc_ids
```
dantuzi: Why do we need this?

rre-dataset-generator/.gitignore
dantuzi: I misunderstood the line. My comment doesn't have any sense

rre-dataset-generator/src/config.py
dantuzi: Let's wait for the introduction of the ruff linter to fix it

rre-dataset-generator/src/writers/quepid_writer.py
```diff
@@ -1,32 +1,37 @@
...
-    @classmethod
-    def build(cls, config: Config, data_store: DataStore):
-        return cls(datastore=data_store)
+    def _get_queries_and_ratings(self, datastore: DataStore) -> List[Tuple[str, str, int]]:
+        """Helper to extract (query_text, doc_id, rating) tuples from the datastore."""
+        result = []
+        for query in datastore.get_queries():
+            for doc_id in datastore.get_doc_ids_for_query(query.id):
+                rating = datastore.get_rating_score(query.id, doc_id)
+                if rating is not None:
```
dantuzi: We can simply use the method `datastore.get_ratings()`

rre-dataset-generator/src/data_store.py
```diff
@@ -0,0 +1,274 @@
...
+        if not self.has_query(rating.query_id):
+            log.warning(f"[add_rating] query_not_found query_id={rating.query_id}")
+            return
+        if not self.has_document(rating.doc_id):
+            log.warning(f"[add_rating] doc_not_found doc_id={rating.doc_id}")
+            return
+
+        key = (rating.query_id, rating.doc_id)
+        if key in self.rating_by_pair:
+            log.debug(f"[add_rating] exists q={rating.query_id} d={rating.doc_id}")
```
dantuzi: warning

rre-dataset-generator/src/data_store.py
```diff
@@ -0,0 +1,274 @@
...
+        if (existing_id := self.query_text_to_query_id.get(key)):
+            log.debug(f"[add_query] exists text='{query.text}' existing_id={existing_id}")
+            return existing_id
+
+        self.queries[query.id] = query
+        self.query_text_to_query_id[key] = query.id
+        log.debug(f"[add_query] added query_id={query.id}")
+        return query.id
+
+    def add_rating(self, rating: Rating) -> None:
```
dantuzi: if it is private, can you please add the `_` before the name?

rre-dataset-generator/src/data_store.py
```diff
@@ -0,0 +1,274 @@
...
+        """Gets all ratings. Complexity: O(P) where P is the number of ratings."""
+        return list(self.rating_by_pair.values())
+
+    def get_ratings_for_query(self, query_id: str) -> List[Rating]:
+        """(MVP) Returns ratings for a query. Complexity: O(P) where P is the number of ratings."""
+        if not self.has_query(query_id):
+            return []
+        return [r for r in self.rating_by_pair.values() if r.query_id == query_id]
+
+    def get_doc_ids_for_query(self, query_id: str) -> List[str]:
```
dantuzi: We don't need this function

rre-dataset-generator/src/data_store.py
dantuzi: This is not needed as well because it is only read in the method no longer need it

rre-dataset-generator/dataset_generator.py
Intrinsical-AI: Done

rre-dataset-generator/src/config.py
Intrinsical-AI: Ok

rre-dataset-generator/src/data_store.py
Intrinsical-AI: Removed `docs_by_query`

rre-dataset-generator/src/writers/quepid_writer.py
Intrinsical-AI: Done

rre-dataset-generator/src/data_store.py
Intrinsical-AI: Ok

rre-dataset-generator/src/data_store.py
Intrinsical-AI: Done

rre-dataset-generator/src/data_store.py
Intrinsical-AI: Done

rre-dataset-generator/dataset_generator.py
```diff
@@ -1,91 +1,102 @@
...
+    """Complete the (query, doc) matrix with LLM scores."""
+    for query_obj in data_store.get_queries():
+        for doc_obj in data_store.get_documents():
+            if not data_store.has_rating_score(query_obj.id, doc_obj.id):
+                score_resp: LLMScoreResponse = llm_service.generate_score(
+                    doc_obj, query_obj.text, config.relevance_scale, config.save_llm_explanation
+                )
+                data_store.create_rating_score(
+                    query_obj.id, doc_obj.id, score_resp.get_score(),
+                    score_resp.get_explanation() if getattr(config, "save_llm_explanation", False) else None
```
nicolo-rinaldi: `get_explaination()` method has been removed

rre-dataset-generator/src/data_store.py
```diff
@@ -0,0 +1,235 @@
...
+        # Primary (id → object)
+        self.docs: Dict[str, Document] = {}
+        self.queries: Dict[str, Query] = {}
+
+        # Ratings storage
+        self.rating_by_pair: Dict[Tuple[str, str], Rating] = {}    # (query_id, doc_id) → Rating 
+
+        # Text based deduplication for queries
+        self.query_text_to_query_id: Dict[str, str] = {}           # query_text → query_id 
+        # TODO: add normalizing function for the text keys (strip, lower, etc).
```
nicolo-rinaldi, line 39: add also the jira ticket if you can

rre-dataset-generator/src/data_store.py
```diff
@@ -0,0 +1,235 @@
...
+    # ────────────────────────────────────────────
+    def add_document(self, doc: Document) -> None:
+        """Adds a document. Complexity: O(1)."""
+        if self.has_document(doc.id):
+            log.warning(f"[add_document] exists doc_id={doc.id}")
+            return
+        self.docs[doc.id] = doc
+        log.debug(f"[add_document] added doc_id={doc.id}")
+
+    def add_query(self, query: Query) -> str:
```
dantuzi: I'm thinking about this implementation to exclude the possibility of having the function caller deal with two query IDs: one returned by this function and one in the query object

```
    def add_query(self, query_text: str) -> Query:
        if (existing_id := self.query_text_to_query_id.get(key)):
            log.warning(f"[add_query] exists text='{query.text}' existing_id={existing_id}")
            return self.queries.get(existing_id)

        query = Query(text=query_text)
        self.queries[query.id] = query
        self.query_text_to_query_id[query_text] = query.id
        log.debug(f"[add_query] added query_id={query.id}")
        return query
```

rre-dataset-generator/dataset_generator.py
```diff
@@ -96,31 +107,33 @@ def retrieve_and_score_documents(config: Config, data_store: DataStore, llm_serv
...
+    # write results
+    output_destination = config.output_destination
+    writer.write(output_destination, data_store)
+    log.info(f"Synthetic Dataset has been generated in: {output_destination}")
 
-    if config.save_llm_explanation:
-        data_store.export_all_records_with_explanation(config.llm_explanation_destination)
-        log.info(f"Dataset with LLM explanation is saved into: {config.llm_explanation_destination}")
+    # save explanation
+    explanation_destination = getattr(config, "save_llm_explanation", False)
```
dantuzi: I don't like this approach, mostly for two reasons:
- The fallback to false if the variable `save_llm_explanation` doesn't exist has to be done only once in the config loader and not every time we use that value
- if the name of `save_llm_explanation` changes, we don't have the support of any tool to detect that possible error

rre-dataset-generator/src/data_store.py
dantuzi, line 39: good idea

rre-dataset-generator/src/data_store.py
```diff
@@ -0,0 +1,235 @@
...
+
+    def has_query(self, query_id: str) -> bool:
+        """Checks for query existence. Complexity: O(1)."""
+        return query_id in self.queries
+
+    def has_rating_score(self, query_id: str, doc_id: str) -> bool:
+        """Checks for a rating by (query, doc) pair. Complexity: O(1)."""
+        return (query_id, doc_id) in self.rating_by_pair
+
+    def has_rating(self, rating_id: str) -> bool:
```
dantuzi: never used

rre-dataset-generator/src/data_store.py
```diff
@@ -0,0 +1,235 @@
...
+    def get_documents(self) -> List[Document]:
+        """Gets all documents. Complexity: O(N) where N is the number of docs."""
+        return list(self.docs.values())
+
+    def get_query(self, query_id: str) -> Optional[Query]:
+        """Gets a single query by its ID, or None if not found."""
+        return self.queries.get(query_id)
+
+    def get_queries(self) -> List[Query]:
+        """Gets all queries. Complexity: O(M) where M is the number of queries."""
```
dantuzi: Please remove this complexity in the comments

rre-dataset-generator/src/data_store.py
```diff
@@ -0,0 +1,235 @@
...
+    def create_rating_score(
+        self, query_id: str, doc_id: str, score: int, explanation: Optional[str] = None
+    ) -> Optional[Rating]:
+        """Create rating (if not exists) and add via `add_rating`. Complexity: O(1)."""
+        if not self.has_query(query_id):
+            log.warning(f"[create_rating_score] query_not_found query_id={query_id}")
+            return None
+        if not self.has_document(doc_id):
+            log.warning(f"[create_rating_score] doc_not_found doc_id={doc_id}")
+            return None
```
dantuzi: This can be removed because the checks are already done in the inner function

rre-dataset-generator/src/data_store.py
```diff
@@ -0,0 +1,235 @@
...
+        tmp_path = self.path.with_name(self.path.name + f".{uuid4().hex}.tmp")
+        tmp_path.write_text(json.dumps(data, indent=2, ensure_ascii=False), encoding=ENCODING)
+        os.replace(tmp_path, self.path)
+        log.info("[save] ok path=%s n_docs=%d n_queries=%d n_ratings=%d",
+                self.path, len(self.docs), len(self.queries), len(self.rating_by_pair))
+
+    def load(self) -> None:
+        if not self.path.exists():
+            return
+        try:
```
dantuzi, line 163: self._clear_all_data() here instead of doing it twice at line 176 and 179

rre-dataset-generator/src/data_store.py
```diff
@@ -0,0 +1,235 @@
...
+
+        # queries
+        for query_as_str in data.get("queries", []):
+            try:
+                self.add_query(Query.model_validate(query_as_str))
+            except ValidationError as e:
+                log.warning("[load] skip_query_invalid data=%s error=%s", query_as_str, e)
+
+        # ratings
+        for rating_as_str in data.get("ratings", []):
```
dantuzi:
```
        for rating_as_str in data.get("ratings", []):
            try:
                robj = Rating.model_validate(rating_as_str)
                self._add_rating(robj)  # verifies refs and creates query→doc link
            except ValidationError as e:
                log.warning("[load] skip_rating_invalid data=%s error=%s", rating_as_str, e)
            
```

nicolo-rinaldi (approved):

rre-dataset-generator/src/model/document.py
```diff
@@ -20,12 +30,23 @@ class Document(BaseModel):
 
     @field_validator('fields')
     @classmethod
-    def check_no_empty_fields(cls, v: Dict[str, Any]) -> Dict[str, Any]:
-        """Validate that the fields dictionary is not empty and its keys are not empty."""
+    def validate_fields(cls, v: Dict[str, Any]) -> Dict[str, Any]:
```
dantuzi: Can you replace this `v` with `field_values`?

rre-dataset-generator/src/model/document.py
```diff
@@ -20,12 +30,23 @@ class Document(BaseModel):
...
+        """Validate that the fields dictionary and its keys are not empty and that all values are JSON-serializable."""
         if not v:
-            log.error('The fields dictionary cannot be empty.')
             raise ValueError('The fields dictionary cannot be empty.')
         if any(not key for key in v.keys()):
-            log.error('Field keys cannot be empty strings.')
             raise ValueError('Field keys cannot be empty strings.')
+
+        def is_jsonable(value: Any) -> bool:
```
dantuzi: is_json_serializable

dantuzi: and please move this function definition to utils

rre-dataset-generator/src/data_store.py
Intrinsical-AI: This change it's contract-breaking, so I had to change few lines invoking this. 

While doing it I realized would be usefull also to optionally pass the ID, so we can keep consistency between the query-id when loading -> Added in the last commit

rre-dataset-generator/dataset_generator.py
Intrinsical-AI: Done. Finally decided that we can let Config.save_llm_explanation Optional[bool] with default == False.

We could also remove Optional, so the value is never "None" (makes sense, since the default it's already a bool), but now should be more robust anyway.

rre-dataset-generator/src/data_store.py
Intrinsical-AI: Ok!

rre-dataset-generator/src/data_store.py
Intrinsical-AI: Removed

rre-dataset-generator/src/data_store.py
Intrinsical-AI: Done :+1:

rre-dataset-generator/src/data_store.py
Intrinsical-AI: Added after the initial check :+1:

rre-dataset-generator/src/model/document.py
Intrinsical-AI: Done! Also moved the helper to the utils file

rre-dataset-generator/src/model/document.py
Intrinsical-AI: Done, moved to utils

rre-dataset-generator/dataset_generator.py
Intrinsical-AI: Done, thanks!

rre-dataset-generator/src/data_store.py
Intrinsical-AI: I'm planning to collect several "small" features/improvements like this, and propose all together

rre-dataset-generator/src/data_store.py
Intrinsical-AI: :+1: Unified with the other methods

rre-dataset-generator/dataset_generator.py
```diff
@@ -1,91 +1,99 @@
...
+
         query_response: LLMQueryResponse = llm_service.generate_queries(doc, num_queries_per_doc)
-        for query_text in query_response.get_queries():
+        for query_ in query_response.get_queries():
             if len(data_store.get_queries()) >= config.num_queries_needed:
                 return
-            query_id: str = data_store.add_query(query_text, doc.id)
-            data_store.add_rating_score(query_id, doc.id, max(config.relevance_label_set),
-                                        "Default max rating is assigned because the query is generated by the document")
+            query_obj: Query = data_store.add_query(query_.text)
```
nicolo-rinaldi: `query_` doesn't have attribute text since is a string

Intrinsical-AI: I had just updated that after running the pipeline :+1:

dantuzi: What does it mean? Why do we need to extract the `.text` from a string?

rre-dataset-generator/src/data_store.py
```diff
@@ -0,0 +1,222 @@
...
+        self.docs[doc.id] = doc
+        log.debug(f"[add_document] added doc_id={doc.id}")
+
+    def add_query(self, query_text_str: str, id: Optional[str] = None) -> Query:
+        """Adds a new query. If text is cached, returns existing Query. If id is given, it's used."""
+        if (existing_id := self.query_text_to_query_id.get(query_text_str)):
+            log.warning(f"[add_query] exists text='{query_text_str}' existing_id={existing_id}")
+            return self.queries[existing_id]
+
+        query = Query(id=id, text=query_text_str) if id else Query(text=query_text_str)
```
dantuzi, line 101: Just out of curiosity, if `id` is None and you still pass to the Query constructor, what's going to happen? Does Pydantic call the default_factory?

dantuzi (approved):

PR #194, opened 2025-08-04
DAGE-22: Add RRE Writer 

nseidan:
Jira issue: https://sease.atlassian.net/browse/DAGE-22

* Added `RreWriter` and a unit test
* Added `index_name`  param to config

Sample `ratings.json`:
```
{
  "index": "testcore",
  "corpora_file": "testcore",
  "id_field": "id",
  "query_placeholder": "$query",
  "query_groups": [
    {
      "name": "What caused the helicopter crash in the Colombian jungle?",
      "queries": [
        {
          "template": "tests/unit/resources/only_q.json",
          "placeholders": {
            "$query": "What caused the helicopter crash in the Colombian jungle?"
          }
        }
      ],
      "relevant_documents": {
        "2": [
          "1"
        ],
        "0": [
          "2",
          "62",
          "147",
          "132",
          "101"
        ]
      }
    },
    {
      "name": "How many soldiers were killed in the helicopter crash in Colombia?",
      "queries": [
        {
          "template": "tests/unit/resources/only_q.json",
          "placeholders": {
            "$query": "How many soldiers were killed in the helicopter crash in Colombia?"
          }
        }
      ],
      "relevant_documents": {
        "2": [
          "1"
        ],
        "0": [
          "2",
          "62",
          "147",
          "132",
          "101"
        ]
      }
    },
  ]
}

```


rre-dataset-generator/dataset_generator.py
```diff
@@ -90,14 +92,20 @@ def add_cartesian_product_scores(llm_service: LLMService, config: Config, data_s
                                                                 endpoint=config.search_engine_collection_endpoint)
     llm: BaseChatModel = LLMServiceFactory.build(LLMConfig.load(config.llm_configuration_file))
     service: LLMService = LLMService(chat_model=llm)
-    writer: AbstractWriter = WriterFactory.build(config.output_format, data_store)
+    writer: AbstractWriter = WriterFactory.build(config.output_format, data_store,
```
Intrinsical-AI: I like this configuration :+1:

rre-dataset-generator/src/writers/rre_writer.py
```diff
@@ -0,0 +1,83 @@
...
+        self.corpora_file = corpora_file
+        self.id_field = id_field
+        self.query_template = query_template
+        self.query_placeholder = query_placeholder
+
+    def _build_json_doc_records(self) -> dict[str, Any]:
+        query_to_doc_ratings = defaultdict(list)
+
+        for query_text, doc_id, rating in self._get_queries_with_ratings():
+            query_to_doc_ratings[query_text].append((doc_id, int(rating)))
```
Intrinsical-AI, line 44: **[Note – No action/reponse required]**

This works as expected. Just a quick note: if 2 queries share the text content -> we're concatenating the results in the same dictionary entry. E.g.:

```
Query 1: (id=1, text="text_1", related_docs = ["1","2"], ratings =[0,2])
Query 2: (id=2, text="text_1", related_docs = ["3","4"], ratings =[1,0])

--> query_to_doc_ratings["text_1"] = [("1", 0), ("2", 2), ("3", 1), ("4", 0)]
```

Just to reminder for the future

rre-dataset-generator/src/writers/writer_factory.py
```diff
@@ -2,17 +2,21 @@
...
+
 class WriterFactory:
     OUTPUT_FORMAT_REGISTRY = {
         "quepid": QuepidWriter,
+        "rre": RreWriter,
     }
 
     @classmethod
-    def build(cls, output_format: str, data_store: DataStore) -> AbstractWriter:
+    def build(cls, output_format: str, data_store: DataStore, **kwargs) -> AbstractWriter:
```
Intrinsical-AI: :+1:

rre-dataset-generator/src/writers/writer_factory.py
```diff
@@ -2,17 +2,21 @@
...
         if output_format not in cls.OUTPUT_FORMAT_REGISTRY:
             log.error("Unsupported output format requested: %s", output_format)
             raise ValueError(f"Unsupported output format: {output_format}")
         log.info("Selected output format: %s", output_format)
-        return cls.OUTPUT_FORMAT_REGISTRY[output_format](data_store)
+        return cls.OUTPUT_FORMAT_REGISTRY[output_format].from_factory(data_store, **kwargs)
```
Intrinsical-AI: :+1:

rre-dataset-generator/tests/unit/resources/good_config.yaml
```diff
@@ -1,5 +1,6 @@
 query_template: "q=#$query##&fq=genre:horror&wt=json"
 search_engine_type: "solr"
+index_name: "mycore"
```
Intrinsical-AI: [NB] Why no /testcore as previously? Different dataset / config files?

rre-dataset-generator/tests/unit/writers/test_rre_writer.py
```diff
@@ -0,0 +1,72 @@
...
+from pathlib import Path
+
+import pytest
+
+from src.config import Config
+from src.search_engine.data_store import DataStore
+from src.writers.rre_writer import RreWriter
+
+
+@pytest.fixture
```
Intrinsical-AI, line 11: :+1:

Intrinsical-AI (approved):
Approved. Added some [NB] comments. 

**_[NB] = Non-Blocking_**

rre-dataset-generator/tests/unit/resources/good_config.yaml
nseidan: testcore or mycore, no difference, i kept mycore cuz it was used initially in this file, so didnt want to change to testcore.  It is a unit test, doesn't make any difference as not reaching the search engine.

rre-dataset-generator/dataset_generator.py
dantuzi: I don't. I think it's more readable and flexible if we pass the config object directly and the builder gets the information it needs

rre-dataset-generator/dataset_generator.py
```diff
@@ -90,14 +92,20 @@ def add_cartesian_product_scores(llm_service: LLMService, config: Config, data_s
...
+                                                 query_template=config.rre_query_template,
+                                                 query_placeholder=config.rre_query_placeholder)
 
     # pipeline starts
     add_user_queries(config, data_store)
 
-    docs_to_generate_queries: List[Document] = search_engine.fetch_for_query_generation(documents_filter=config.documents_filter,
-                                                                                        doc_number=config.doc_number,
-                                                                                        doc_fields=config.doc_fields)
+    docs_to_generate_queries: List[Document] = search_engine.fetch_for_query_generation(
```
dantuzi: same here

rre-dataset-generator/src/config.py
```diff
@@ -65,15 +66,27 @@ def relevance_label_set(self) -> set[int]:
...
             raw_config = yaml.safe_load(f)
 
         log.debug("Configuration file loaded successfully.")
+
+        if raw_config.get("output_format") == "rre":
+            return RreConfig(**raw_config)
         return cls(**raw_config)
+
+
+class RreConfig(Config):
```
dantuzi: Why do we need an extra config file?

rre-dataset-generator/src/config.py
```diff
@@ -65,15 +66,27 @@ def relevance_label_set(self) -> set[int]:
...
+    output_format: Literal['rre'] = 'rre'
+    corpora_file: FilePath = Field(..., description="JSON formatted dataset file.")
+    id_field: str = Field(..., description="ID field for the unique key.")
+    rre_query_template: FilePath = Field(..., description="Query template for rre evaluator.")
```
dantuzi: Do we need the query template for the RRE exporter? Can't we use the query template we have already? They must be the same

rre-dataset-generator/src/writers/rre_writer.py
```diff
@@ -0,0 +1,83 @@
...
+log = logging.getLogger(__name__)
+
+
+class RreWriter(AbstractWriter):
+    """
+    Writes query ratings in RRE format (ratings.json).
+    """
+
+    @classmethod
+    def from_factory(cls, data_store, **kwargs):
```
dantuzi: How about something like "build"?

rre-dataset-generator/src/config.py
nseidan: we need extra params to pass to RRE exporter such as 
corpora_file  path, id_field name,  query_placeholder, query_template for ratongs.json file generation

rre-dataset-generator/dataset_generator.py
nseidan: well, if we want to pass config as an arg here, we need to refactor many classes, as `fetch_for_query_generation` is from base class. I would suggest to create a jira issue to do the refactoring. What do you think?

rre-dataset-generator/dataset_generator.py
nseidan: updated

rre-dataset-generator/src/config.py
nseidan: we need extra params to pass to RRE exporter such as
corpora_file path, id_field name, query_placeholder, query_template for ratings.json file generation

rre-dataset-generator/src/config.py
nseidan: here we need to provide file path for that query template in json format, for example, 
only_q.json: 
`{"q" :  "$query"}` 

but for  query template it is a text if it is solr, key-value pairs if it is opensearch/elasticsearch, and  vespa has a different type; so making it here  user provides that file, instead of parsing the given user query template on different cases and create the json file.

rre-dataset-generator/src/writers/rre_writer.py
nseidan: updated

rre-dataset-generator/src/config.py
dantuzi: I mean, why can't we add them to the "normal" config file?
I guess it's more maintainable if we keep all config variables in a single file
Maybe in the config file example, we can write something like that

```yaml
### The following section is read only if we want to export to RRE
# corpora_file path = "XXX"
# id_field_name = "YYY"
# query_placeholder = "ZZZ"
```

Don't you think this is simpler?

rre-dataset-generator/src/config.py
dantuzi: I still think we should have a single parameter for the query template because in the future, RRE should also work with OpenSearch and Vespa.
I'll resolve this conversation for now, but this is something we need to standardize afterwards

dantuzi: Created a ticket in the backlog for the future: https://sease.atlassian.net/browse/DAGE-43

rre-dataset-generator/src/config.py
nseidan: updated

dantuzi (approved):

PR #195, opened 2025-08-11
DAGE-41:[DRAFT] Implement MTEB Custom Tasks

nseidan:
Jira issue:
https://sease.atlassian.net/browse/DAGE-41

* Added 2 custom tasks for retrieval and reranking
* Added data samples : corpus, queries, candidates, and relevant docs

Limitation:
Built-in cache wrapper puts all embeddings coming from queries and docs into the same place.  

----
Code:

Abstract class for Reranking Task: https://github.com/embeddings-benchmark/mteb/blob/main/mteb/abstasks/AbsTaskReranking.py

Abstract class for Retrieval Task: https://github.com/embeddings-benchmark/mteb/blob/main/mteb/abstasks/AbsTaskRetrieval.py

Cache Wrapper: https://github.com/embeddings-benchmark/mteb/blob/main/mteb/models/cache_wrapper.py


rre-dataset-generator/src/mteb/custom_retrieval_task.py
```diff
@@ -0,0 +1,46 @@
...
+        domains=["Engineering"],
+        task_subtypes=["Article retrieval"],
+        license="not specified",
+        annotations_creators="derived",
+        sample_creation="created",
+        dataset={
+            "name": "my_local_dataset",
+            "path": "local",
+            "revision": "v1",
+            "url": "https://github.com/SeaseLtd/rated-ranking-evaluator"
```
Intrinsical-AI, line 32: **[Question]**

Same as in the Reranker class?

rre-dataset-generator/src/mteb/custom_reranking_task.py
```diff
@@ -0,0 +1,69 @@
...
+        type="Reranking",
+        category="s2p",
+        eval_splits=["test"],
+        eval_langs=["en"],
+        main_score="map",
+        dataset={
+            "name": "my_local_dataset",
+            "path": "local",
+            "revision": "v1",
+            "url": "https://github.com/SeaseLtd/rated-ranking-evaluator"
```
Intrinsical-AI, line 54: **[Question]**

I see you’re loading the data with the correct `path` in the other function, but I have some doubts about how `url` and `path` are set. From what I checked:

```
path → local path (absolute or relative) to the folder containing the dataset structure  
url  → remote location to fetch the dataset from if not available locally  
```

If that’s correct, would it make sense to use something like:

```json
"path": "/rre-dataset-generator/data",
"url": "https://github.com/SeaseLtd/rated-ranking-evaluator/tree/main/rre-dataset-generator/data"
```

?

rre-dataset-generator/src/mteb/main.py
```diff
@@ -0,0 +1,37 @@
...
+    print(len(vectors))  # combined corpus + queries 'text' fields
+    for vector in vectors:
+        print(vector)
+
+    cached.close()
+    del cached
+
+    vector_map.close()
+    del vector_map
+
```
Intrinsical-AI: **[Suggestion]**

Add a simple unit test to check CachedEmbeddingWrapper / TextVectorMap and/or Tasks?

Intrinsical-AI (commented):
**[Minor]** Questions relating to paths.


**[Non-blocking]** Suggested adding unit tests

rre-dataset-generator/src/mteb/custom_retrieval_task.py
nseidan: it's just a sample.

Intrinsical-AI: Oh, I checked and you're right. As implemented it's just used as metadata information and we perform the real loading on the load_dataset() function. 

It's not mandatory but makes sense align MyRetrievalTask.metadata.dataset values, (path,url) with the routes used in `load_dataset`, instead DATA_ROOT as a fixed:

E.g
```
def load_data(self, ..):
        dataset = datasets.load_dataset(path=self.metadata.dataset["path"],...)
```

rre-dataset-generator/src/mteb/main.py
nseidan: this code won't be merged to main; it is just for verifiying my MTEB investigation. As a follow-up of this, would be to create actual tasks and implement. So I don't need to write unit tests as part of this.

rre-dataset-generator/src/mteb/custom_reranking_task.py
nseidan: again, this is not real implementation, it is just for verification as MTEB custom tasks work.

Intrinsical-AI:
I agree on the approach. In my opinion with few small adaptations could be merged in the project as prod code

Intrinsical-AI (approved):
I agree on the approach, with few small adaptations could be merged in the production branch in my opinion.

dantuzi (approved):
This is a temporary code just to validate how MTEB works

nseidan:
I closed this PR, this should not be merged.

PR #198, opened 2025-08-14
DAGE 15: integration test

nicolo-rinaldi:
What i did in this ticket is to create a test file for all the search engines supported up to now (solr, elasticsearch and opensearch) → the only way to test in one go is to create one new docker-compose file (called docker-compose.full.ymlthat is used to extend all the other (in a future vespa integration, the vespa extension must be added). For each search engine there are 4 basic test function + one test for the full pipeline, that is contained in the file end_to_end_pipeline.py. It makes sense to move this test in a e2e/ folder inside tests/ but is going to be done (if needed) later on. 

For opensearch, I changed the port (9200 → 9300) since there were ports conflict due to the unification of the Docker container.

More than this, also a test with the interaction between the datastore and the writers has been added to the tests/integration/ folder. 

If the DataStore refactor on ticket [DAGE-37](https://sease.atlassian.net/browse/DAGE-37) is going to be merged before this, all the code that involves DataStore class must be changed (I think is only inside the latter tests I mentioned. On the opposite, there will be more code to solve conflict for Pablo in the Refactor.

[ticket](https://sease.atlassian.net/browse/DAGE-15)


rre-dataset-generator/tests/conftest.py
```diff
@@ -0,0 +1,23 @@
+import os
+import subprocess
+import pytest
+
+@pytest.fixture(scope="session")
```
Intrinsical-AI, line 5: [Suggestion]

The conftest it's loaded for all the test files within that directoy. I think we could move this file to test/integration/ to avoid loading this when executing the unit (and the e2e in the future if we add it)

rre-dataset-generator/tests/integration/docker-compose.full.yml
```diff
@@ -0,0 +1,36 @@
...
+
+  solr:
+    extends:
+      file: docker-compose.solr.yml
+      service: solr
+
+  solr-init:
+    extends:
+      file: docker-compose.solr.yml
+      service: solr-init
```
Intrinsical-AI, line 36: Very clear and readable :+1:

rre-dataset-generator/tests/integration/end_to_end_pipeline.py
```diff
@@ -0,0 +1,95 @@
...
+from src.model.document import Document
+from src.model.query_rating_context import QueryRatingContext
+from src.search_engine.data_store import DataStore
+from src.search_engine.search_engine_base import BaseSearchEngine
+from src.search_engine.search_engine_factory import SearchEngineFactory
+from src.writers.abstract_writer import AbstractWriter
+from src.writers.writer_factory import WriterFactory
+
+
+def end_to_end_pipeline_with_llm_mock(config, tmp_path):
```
Intrinsical-AI, line 24: [Notes]

1. If the file or the method does not start with "test_" pytest doesn't recognize it by default, so this has to be manually triggered (I think it's the intended behaviour)
2. I think is more like an e2e than integration test, but looks solid anw

Intrinsical-AI (approved):

PR #200, opened 2025-08-15
DAGE-47: Add MtebWriter

nseidan:
* jira: [jira](https://sease.atlassian.net/browse/DAGE-47)
* added MTEB writer 
* under `output_destination` dir, the following three files will be saved: `data/corpus.jsonl`,  `data/queries.jsonl`, `data/candidates.jsonl`
* added a unit test


rre-dataset-generator/src/writers/mteb_writer.py
```diff
@@ -0,0 +1,79 @@
...
+            for query_context in self.datastore.get_queries():
+                query_id = query_context.get_query_id()
+                for doc_id in query_context.get_doc_ids():
+                    if query_context.has_rating_score(doc_id):
+                        rating_score = query_context.get_rating_score(doc_id)
+
+                        row = {"query_id": query_id, "doc_id": doc_id, "rating": rating_score}
+                        file.write(json.dumps(row, ensure_ascii=False) + "\n")
+
+    def write(self, output_path: str | Path) -> None:
```
nicolo-rinaldi, line 71: **[IDEA]**
we could also use this to write all 3 files in one go

nicolo-rinaldi (approved):

rre-dataset-generator/src/writers/mteb_writer.py
nseidan: adjusted

rre-dataset-generator/src/writers/mteb_writer.py
```diff
@@ -0,0 +1,95 @@
...
+                row = {"id": query_id, "text": query_text}
+                file.write(json.dumps(row, ensure_ascii=False) + "\n")
+
+    def _write_candidates(self, output_path: str | Path) -> None:
+        """
+        Writes candidates to JSONL file:
+        {"query_id": <query_id>, "doc_id": <doc_id>, "rating": <rating_score>}
+        """
+        path = Path(output_path)
+        os.makedirs(path.parent, exist_ok=True)
```
dantuzi: This is already done in the method `write`, so we get the output_path with type `Path` and avoid creating the directory.
If the method were public as it was before, this piece of code would be necessary but now the method is private

nseidan: updated

dantuzi (approved):

PR #201, opened 2025-08-19
DAGE-31: M3 index doc to solr with embeddings

nicolo-rinaldi:
Jira ticket: https://sease.atlassian.net/browse/DAGE-31

Updated the docker-related files inside solr-init/ folder to deal with the file output/embeddings/doc_embeddings.jsonland index documents’ embeddings into the solr instance. For now the embedding dimension is fixed to 10, then it needs to be changed later on when we have a complete configuration file. The folder where the file containing the embeddings is fixed. If the folder is not found, then an image creates the folder needed but empty. In this way, the code works even if the file doesn’t exists, it just index the documents without any embedding.

### DIFF
--- a/rre-dataset-generator/tests/integration/solr-init/solr-init.sh
+++ b/rre-dataset-generator/tests/integration/solr-init/solr-init.sh
@@ -57,11 +57,11 @@ else
 fi
 
 if [ "$COUNT_FOUND" -eq 0 ]; then
-  echo "[INFO] Indexing into Solr…"
+  echo "[INFO] Indexing dataset…";
   curl -f -X POST \
        -H "Content-Type: application/json" \
        --data-binary @"$TMP_FILE" \
-       "$COLLECTION_ENDPOINT/update?commit=true"
+       $COLLECTION_ENDPOINT/update?commit=true;
   echo "[INFO] Done indexing.";
 else
   echo "[INFO] Collection already contains data, skipping indexing.";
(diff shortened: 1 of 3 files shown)


nseidan (approved):

rre-dataset-generator/tests/integration/solr-init/solr-init.sh
```diff
@@ -24,11 +27,42 @@ SOLR_RESPONSE=$(curl -sf $COLLECTION_ENDPOINT/select\?q\=\*:\*)
...
+if [ -f "$EMBEDDINGS_FILE" ] && command -v jq >/dev/null 2>&1; then
+  VECTOR_DIM=$(head -n 1 "$EMBEDDINGS_FILE" | jq '.vector | length')
+  echo "[INFO] Embeddings file found -> detected embedding dimension: $VECTOR_DIM"
+  echo "[INFO] jq available → merging…"
+  jq --slurpfile emb "$EMBEDDINGS_FILE" '
+    map(. as $d | $emb[] | select(.id == $d.id) | $d + {vector: .vector}) as $with_vec
+    | map(if ($with_vec[]? | select(.id == .id) | .id) == .id
+          then ($with_vec[] | select(.id == .id))
+          else . end)
+  ' "$DATASET_FILE" > "$TMP_FILE"
```
dantuzi: The current script doesn't work. Please replace it with the follow

```bash
jq --slurpfile emb "$EMBEDDINGS_FILE" '
  map(
    . as $doc
    | ($emb[] | select(.id == $doc.id)) as $e
    | if $e != null
      then $doc + {vector: $e.vector}
      else $doc
      end
  )
' "$DATASET_FILE" > "$TMP_FILE"
```

dantuzi (approved):

PR #202, opened 2025-08-20
DAGE-48: (M2) Add MTEB conf file

nseidan:
Jira issue: https://sease.atlassian.net/browse/DAGE-48
Added: mteb config file and unit test


nicolo-rinaldi (approved):
Compliant to what we did on the M1 (dataset-generator)

dantuzi (approved):

PR #204, opened 2025-08-27
[bugfix] rre + quepid path bugfix

nicolo-rinaldi:
Changes to tests + writers class (RRE and Quepid) according on how config file is written

No ticket, just a bugfix

### DIFF
--- a/rre-dataset-generator/src/writers/quepid_writer.py
+++ b/rre-dataset-generator/src/writers/quepid_writer.py
@@ -25,7 +25,7 @@ class QuepidWriter(AbstractWriter):
 
     def write(self, output_path: str | Path, datastore: DataStore) -> None:
         """Writes queries and their scored documents to a CSV file in Quepid format."""
-        output_path = Path(output_path)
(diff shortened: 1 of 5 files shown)


rre-dataset-generator/src/writers/quepid_writer.py
```diff
@@ -25,7 +25,7 @@ def _get_queries_and_ratings(self, datastore: DataStore) -> List[Tuple[str, str,
 
     def write(self, output_path: str | Path, datastore: DataStore) -> None:
         """Writes queries and their scored documents to a CSV file in Quepid format."""
-        output_path = Path(output_path)
+        output_path = Path(output_path) / "quepid.csv"
```
Intrinsical-AI: I agree, only mention we're fixing the path, so we won't be able to have 2 distinct quepid files, without manual re-naming :+1:

Intrinsical-AI (approved):

dantuzi (approved):
Created a constant to avoid replicating the same file name multiple times

PR #205, opened 2025-08-28
Dage 44: Introduce Mypy

Intrinsical-AI:
### [Task: DAGE-44](https://sease.atlassian.net/browse/DAGE-44?atlOrigin=eyJpIjoiZGFmNmViODJlNDAzNDI2NGJhNGYxYjdmNmQwNWQ4ZTEiLCJwIjoiaiJ9)

- Removed __ init __ from rated-ranked-evaluator/rre-dataset-generator: it was causing errors. The package it's the content inside /src/, not the whole folder
- Added mypy with basic but robust configuration.
- Updated config.py to avoid mypy confusions with optionals
- Unified setup + conf on the pyproject.toml
- Removed __ init __ files from tests (pytest does not expect packages + we avoid some mypy confusions)

### DIFF
--- a/rre-dataset-generator/pyproject.toml
+++ b/rre-dataset-generator/pyproject.toml
@@ -32,3 +32,37 @@ dev = [
 testpaths = ["tests"]
 pythonpath = ["."]
 addopts = "-v --tb=short"
+
+[tool.mypy]
+python_version = "3.10"
+plugins = ["pydantic.mypy"]
+files = ["src", "dataset_generator.py"]
+exclude = "^tests/integration/"
+# Strict but reasonable
+disallow_untyped_defs = true
+check_untyped_defs = true
+no_implicit_optional = true
+warn_return_any = true
+show_error_codes = true
+pretty = true
+
+# Silence libs without stubs
+[[tool.mypy.overrides]]
+module = [
+    "langchain.*",
+    "langchain_core.*",
+    "langchain_openai.*",
+    "langchain_google_genai.*",
+    "dotenv.*"
+]
+ignore_missing_imports = true
+
+# Allow untyped functions in tests for speed
+[[tool.mypy.overrides]]
+module = "tests.*"
+disallow_untyped_defs = false
+check_untyped_defs = false
+
+[tool.pydantic-mypy]
+init_typed = true
+warn_required_dynamic_aliases = true
--- a/rre-dataset-generator/pyproject.toml
+++ b/rre-dataset-generator/pyproject.toml
@@ -5,7 +5,6 @@ description = "Dataset Generator project"
 readme = "README.md"
 requires-python = ">=3.10"
 dependencies = [
-    "argparse>=1.4.0",
     "pydantic>=2.0",
     "pytest>=8.4.1",
     "pyyaml>=6.0.2",
@@ -15,3 +14,21 @@ dependencies = [
     "langchain-google-genai>=2.1.7",
     "python-dotenv>=1.1.1",
 ]
+
+[dependency-groups]
+dev = [
+    "mypy>=1.17.1",
+    "mypy-extensions>=1.1.0",
+    "pre-commit>=4.3.0",
+    "pytest>=8.4.1",
+    "pytest-cov>=6.2.1",
+    "ruff>=0.12.10",
+    "typing-extensions>=4.14.1",
+    "types-PyYAML>=6.0.2",
+    "types-requests>=2.32.4",
+]
+
+[tool.pytest.ini_options]
+testpaths = ["tests"]
+pythonpath = ["."]
+addopts = "-v --tb=short"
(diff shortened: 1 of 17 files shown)


rre-dataset-generator/src/search_engine/elasticsearch_search_engine.py
```diff
@@ -19,8 +19,8 @@ class ElasticsearchSearchEngine(BaseSearchEngine):
     """
     Elasticsearch implementation to search into a given collection
     """
-    def __init__(self, endpoint: HttpUrl | str):
-        super().__init__(endpoint)
+    def __init__(self, endpoint: Union[HttpUrl, str]):
```
nicolo-rinaldi: since we used `|` instead of `Union` everywhere, I'd keep `|` also in this case

rre-dataset-generator/src/writers/abstract_writer.py
nicolo-rinaldi, line 1: maybe you can change also this file according to the changes done for `BaseSearchEngine` class (the `NotImplementedError` way)

rre-dataset-generator/pyproject.toml
nicolo-rinaldi, line 1: just out of curiosity: this new lines are substituting the file `mypy.ini`? In this case, I think is a good idea to align also `rre-emebddings` to this working pattern

Intrinsical-AI: Yes it is!

 Here I did it simply for simplicity + reduce the number of files. In both mypy already works (very similar in effect and usage) so I don't see clear gains short-term. I'd consider focusing on adding ruff and CI first (pending on the backlog), on each project. Anw, if you think it's critical, feel free to propose :+1:

Intrinsical-AI: Yes it is!

 Here I did it simply for simplicity + reduce the number of files. In both mypy already works (very similar in effect and usage) so I don't see clear gains short-term. I'd consider focusing on adding ruff and CI first (pending on the backlog), on each project. Anw, if you think it's critical, feel free to propose :+1:

rre-dataset-generator/src/writers/abstract_writer.py
Intrinsical-AI: Good idea. Makes sense. Comming next.

 Also: re-adding the ABC heritage in the BaseSearchEngine for consistency

rre-dataset-generator/src/search_engine/elasticsearch_search_engine.py
Intrinsical-AI: You're right, | it's the stardard way. 

Btw: just checked the rest of the class and this it's the only to allow an str. The rest forces `HttpUrl`, so I will unify it

nicolo-rinaldi (approved):

rre-dataset-generator/src/config.py
```diff
@@ -27,31 +27,31 @@ class Config(BaseModel):
...
+    corpora_file: Optional[FilePath] = Field(None, description="JSON formatted dataset file.")
+    id_field: Optional[str] = Field(None, description="ID field for the unique key.")
+    rre_query_template: Optional[FilePath] = Field(None, description="Query template for rre evaluator.")
+    rre_query_placeholder: Optional[str] = Field(None, description="Key-value pair to substitute in the rre query template.")
     verbose: bool = False
 
 
     @field_validator('doc_fields')
-    def check_no_empty_fields(cls, v):
+    def check_no_empty_fields(cls, v: List[str]) -> List[str]:
```
dantuzi: We should fix the variable name `v`

rre-dataset-generator/src/writers/writer_factory.py
```diff
@@ -15,7 +17,7 @@ class WriterFactory:
     }
 
     @classmethod
-    def build(cls, output_format: str) -> AbstractWriter:
+    def build(cls, output_format: str) -> Any:
```
dantuzi: Why `any`? Is it because this method doesn't return the object `AbstractWriter` but an instance of it? I saw you previously used Type[class name]; it probably works here too

Intrinsical-AI: I was getting this error with mypy checks:

```
src/writers/writer_factory.py:25: error: Returning Any from function declared to return "AbstractWriter"  [no-any-return]
            return cls.OUTPUT_FORMAT_REGISTRY[output_format]()
            ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
Found 1 error in 1 file (checked 28 source files)
```

So I switched to -> Any only to bypass mypy’s no-any-return, but that was the wrong fix. The real issue is the "OUTPUT_FORMAT_REGISTRY" being untyped. 

Comming next!

rre-dataset-generator/src/config.py
Intrinsical-AI: We have multiple singe-char variables along the code. I think ruff marks it, so I'd try to get that working instead of "manual" search. Anw: I'll update manually at least the config.py ones for this PR.
 :+1:

dantuzi (approved):

PR #206, opened 2025-08-28
DAGE-50: (M2) Add MTEB embeddings

nseidan:
Jira: https://sease.atlassian.net/browse/DAGE-50
* Added embeddings writer to write embeddings into `documents_embeddings.jsonl` and `queries_embeddings.jsonl` 
* Added unit test


rre-embeddings/tests/unit/test_embedding_writer.py
```diff
@@ -0,0 +1,79 @@
...
+    )
+
+    writer.write(config.embeddings_dest)
+
+    embedding_dir = Path("output/embeddings")
+    docs_file = embedding_dir / "documents_embeddings.jsonl"
+    queries_file = embedding_dir / "queries_embeddings.jsonl"
+    assert docs_file.exists()
+    assert queries_file.exists()
+
```
nicolo-rinaldi, line 74: Does this overwrites some possible already computed files in the folder `output/embeddings/`? If so, I think we need to find a solution to avoid this

nseidan: in real scenario no, as cache is a persistent folder, and embeddings will be imported from that folder. But in this test case, it is not, because of mock CacheEmbeddingWrapper. I used mock CacheEmbeddingWrapper to behave like a cache with fake  static vectors returned.  Also this is a unit test, not integration.

nseidan: updated

nicolo-rinaldi (approved):

dantuzi (approved):

PR #207, opened 2025-09-01
DAGE-45: Added ruff to dataset-generator code + fixed the code

nicolo-rinaldi:
DAGE 45: [Jira ticket](https://sease.atlassian.net/browse/DAGE-45)

Changed also readme to add bash scripts for mypy and ruff checks.

### DIFF
--- a/rre-dataset-generator/dataset_generator.py
+++ b/rre-dataset-generator/dataset_generator.py
@@ -74,7 +74,7 @@ def add_cartesian_product_scores(config: Config, data_store: DataStore, llm_serv
                 )
 
 
-def expand_docset_with_search_engine_topK(config: Config, data_store: DataStore,
+def expand_docset_with_search_engine_top_k(config: Config, data_store: DataStore,
                                  llm_service: LLMService, search_engine: BaseSearchEngine) -> None:
     """Retrieve docs for each query and score the (q, doc) pairs."""
     for query_obj in data_store.get_queries():
@@ -100,7 +100,7 @@ if __name__ == "__main__":
     log: Logger = get_and_setup_logging(args.verbose)
 
     # setup
-    data_store: DataStore = DataStore()
+    data_store: DataStore = DataStore(ignore_saved_data=True)
     search_engine: BaseSearchEngine = SearchEngineFactory.build(
         search_engine_type=config.search_engine_type,
         endpoint=config.search_engine_collection_endpoint
@@ -119,7 +119,7 @@ if __name__ == "__main__":
     add_cartesian_product_scores(config, data_store, service)
 
     # expand the docset with search engine topK (adding direct ratings)
-    expand_docset_with_search_engine_topK(config, data_store, service, search_engine)
+    expand_docset_with_search_engine_top_k(config, data_store, service, search_engine)
 
     # write results
     output_destination = config.output_destination
--- a/rre-dataset-generator/dataset_generator.py
+++ b/rre-dataset-generator/dataset_generator.py
@@ -9,7 +9,7 @@ from src.utils import parse_args
 from src.logger import configure_logging
 from src.llm import LLMConfig, LLMService, LLMServiceFactory
 from src.model import Document, Query,  LLMQueryResponse, LLMScoreResponse
-from src.writers import WriterFactory, AbstractWriter, MtebWriter, QuepidWriter
+from src.writers import WriterFactory, AbstractWriter
 from src.search_engine import SearchEngineFactory, BaseSearchEngine
 from src.data_store import DataStore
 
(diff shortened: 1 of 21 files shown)


rre-dataset-generator/ruff.toml
```diff
@@ -0,0 +1,77 @@
...
+indent-width = 4
+
+# Assume Python 3.9
+target-version = "py39"
+
+[lint]
+# Enable Pyflakes (`F`) and a subset of the pycodestyle (`E`) codes by default.
+# Unlike Flake8, Ruff doesn't enable pycodestyle warnings (`W`) or
+# McCabe complexity (`C901`) by default.
+select = ["E4", "E7", "E9", "F"] # "I", "N"
```
Intrinsical-AI: **[Request]**

If you add "N" you get a lot of errors? I ask because talking with Daniele we assumed that ruff checked the variable names. Adding "N", would be nice if possible.

To check variable naming lenght (and avoid single-letter naming por example), we could add:
```
[lint.flake8-naming]
variable-rgx = "^(i|j|k|[a-z_][a-z0-9_]{2,30})$
```

rre-dataset-generator/src/llm/llm_service.py
```diff
@@ -70,16 +70,16 @@ def generate_score(self, document: Document, query: str, relevance_scale: str,
 
         if explanation:
             system_prompt += (
-                f"Return ONLY a **valid JSON** object with two keys:"
+                "Return ONLY a **valid JSON** object with two keys:"
                 " `score`: the related score as an integer value\n"
                 " `explanation`: your concise explanation for that score\n"
-                f"As an example, I expect a JSON response like the following: "
-                f"{{\"score\": \"integer value\",\"explanation\": \"I rated this score because...\" }}"
+                "As an example, I expect a JSON response like the following: "
```
Intrinsical-AI, line 76: **[Non-blockin]**

I'd small-refactor this: we can share a base string template for both cases, and inyect the example string based on the case forr example. But this is not from this PR, only a suggestion for the future.

rre-dataset-generator/ruff.toml
```diff
@@ -0,0 +1,77 @@
...
+    ".vscode",
+    "__pypackages__",
+    "_build",
+    "buck-out",
+    "build",
+    "dist",
+    "node_modules",
+    "site-packages",
+    "venv",
+]
```
Intrinsical-AI, line 29: **[Idea]**

For the future: I just think that could be useful to dump all the .gitginore files here! (and in Mypy).

rre-dataset-generator/ruff.toml
```diff
@@ -0,0 +1,77 @@
...
+# McCabe complexity (`C901`) by default.
+select = ["E4", "E7", "E9", "F"] # "I", "N"
+ignore = []
+
+# Allow fix for all enabled rules (when `--fix`) is provided.
+fixable = ["ALL"]
+unfixable = []
+
+# Allow unused variables when underscore-prefixed.
+dummy-variable-rgx = "^(_+|(_+[a-zA-Z0-9_]*[a-zA-Z0-9]+?))$"
```
Intrinsical-AI, line 50: **[Question]**
Are we sure we want to keep unused variables when starting with underscore?

Intrinsical-AI (approved):
None critical comments, some ideas / suggestions

Intrinsical-AI:
:+1: Nice!

rre-dataset-generator/ruff.toml
nicolo-rinaldi: This is useful for tests. When we are raising errors we are just calling some functions and we are using the underscore to name it

rre-dataset-generator/dataset_generator.py
```diff
@@ -100,7 +100,7 @@ def expand_docset_with_search_engine_topK(config: Config, data_store: DataStore,
     log: Logger = get_and_setup_logging(args.verbose)
 
     # setup
-    data_store: DataStore = DataStore()
+    data_store: DataStore = DataStore(ignore_saved_data=True)
```
dantuzi: Why? If we ignore the saved data, when the script starts, it doesn't "recycle" the work done previously.
In other words, if we have a failure during the process and the script crashes, when we restart it, we want to re-use the partial data calculated before.
Am I missing something?

rre-dataset-generator/src/config.py
```diff
@@ -37,20 +37,23 @@ class Config(BaseModel):
 
 
     @field_validator('doc_fields')
+    @classmethod
```
dantuzi, line 40: Just out of curiosity, what error did Ruff give here?

nicolo-rinaldi: "he" was complaining that there was `cls` instead of `self` inside the definition of the function

rre-dataset-generator/dataset_generator.py
nicolo-rinaldi: changed

dantuzi (approved):

PR #209, opened 2025-09-02
Dage 46: Linters as PR check

Intrinsical-AI:
TASK: [Dage 46: Linters as PR check](https://sease.atlassian.net/browse/DAGE-46?atlOrigin=eyJpIjoiM2Q2YzNiNTkzYmE2NDJjNDlkY2NhMzAwMzkxYzg4Y2YiLCJwIjoiaiJ9)

Changed the CI current workflow  (test) to add linting checks with mypy and ruff on pushes / PRs to dataset-generator branch.

- Parallelized ruff + mypy + pytest via uv 
- added.py files filter to trigger workflow
- Shared cache, lockfile.
- Avoided concurrency and check duplications

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -2,9 +2,16 @@ name: CI Pipeline
 
 on:
   pull_request:
-    branches: [main, dataset-generator]
+    types: [opened, synchronize, reopened] # conditions on PR to apply checks 
+    # branches: [] -  ommited. Apply checks on all PRs against any branch when types are met
     paths:
       - 'rre-tools/**/*.py'
+      - 'rre-tools/pyproject.toml'
+      - 'rre-tools/uv.lock'
+      - 'rre-tools/ruff.toml'
+      - 'rre-tools/mypy.ini'
+      - 'rre-tools/pytest.ini'
+      - 'rre-tools/tests/**'
       - '.github/workflows/**'
 
 concurrency:
@@ -13,7 +20,6 @@ concurrency:
 
 permissions:
   contents: read
-  pull-requests: write # for annotations in the PR UI (e.g:mypy, ruff)
 
 jobs:
   lint:
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -1,31 +1,104 @@
-name: Unit Tests
+name: CI Pipeline
 
 on:
   pull_request:
-    branches: [ main, dataset-generator ]
+    branches: [main, dataset-generator]
+    paths:
+      - 'rre-tools/**/*.py'
+      - '.github/workflows/**'
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.sha }}
+  cancel-in-progress: true # cancel older runs when pushing several times to the same PR
+
+permissions:
+  contents: read
+  pull-requests: write # for annotations in the PR UI (e.g:mypy, ruff)
 
 jobs:
-  test:
+  lint:
+    name: Linting (ruff)
     runs-on: ubuntu-latest
-
+    timeout-minutes: 15
+    defaults:
+      run:
+        working-directory: rre-tools
     steps:
-    - name: Checkout code
-      uses: actions/checkout@v4
+      - uses: actions/checkout@v4
+      - uses: actions/setup-python@v5
+        with:
+          python-version: "3.11"
+      - uses: astral-sh/setup-uv@v6
+        with:
+          version: "latest"
+      - name: Cache uv and venv
+        uses: actions/cache@v4
+        with:
+          path: |
+            ~/.cache/uv
+            rre-tools/.venv
+          key: ${{ runner.os }}-uv-${{ hashFiles('rre-tools/uv.lock') }} # Common cache: key = hash OS + deps
+          restore-keys: |
+            ${{ runner.os }}-uv-
+      - name: Install dependencies (locked)
+        run: uv sync --locked
+      - name: Run ruff check
+        run: uv run ruff check --output-format=github --force-exclude . # exclude routes contained in .gitignore / .ruffignore
 
-    - name: Set up Python
-      uses: actions/setup-python@v4
-      with:
-        python-version: "3.11"
+  type-check:
+    name: Type Checking (mypy)
+    runs-on: ubuntu-latest
+    timeout-minutes: 15
+    defaults:
+      run:
+        working-directory: rre-tools
+    steps:
+      - uses: actions/checkout@v4
+      - uses: actions/setup-python@v5
+        with:
+          python-version: "3.11"
+      - uses: astral-sh/setup-uv@v6
+        with:
+          version: "latest"
+      - name: Cache uv and venv
+        uses: actions/cache@v4
+        with:
+          path: |
+            ~/.cache/uv
+            rre-tools/.venv
+          key: ${{ runner.os }}-uv-${{ hashFiles('rre-tools/uv.lock') }} # Common cache: key = hash OS + deps
+          restore-keys: |
+            ${{ runner.os }}-uv-
+      - name: Install dependencies (locked)
+        run: uv sync --locked
+      - name: Run mypy
+        run: uv run mypy src/ --install-types --non-interactive # avoid missing stubs
 
-    - name: Install uv
-      uses: astral-sh/setup-uv@v3
-      with:
-        version: "latest"
-
-    - name: Install dependencies
-      run: uv sync
-      working-directory: rre-tools
-
-    - name: Run unit tests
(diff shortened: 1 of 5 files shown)


.github/workflows/test.yml
```diff
@@ -1,13 +1,68 @@
...
 
 on:
   pull_request:
-    branches: [ main, dataset-generator ]
+    branches: [main, dataset-generator]
+  push:
+    branches: [main, dataset-generator]
+    paths:
+      - 'rre-dataset-generator/**/*.py'
+      - '!rre-dataset-generator/**/__pycache__/**'
```
dantuzi: Why? This triggers the execution of all jobs when the PR gets merged

.github/workflows/test.yml
```diff
@@ -1,13 +1,68 @@
...
+        python-version: "3.11"
+
+    - name: Install uv
+      uses: astral-sh/setup-uv@v3
+      with:
+        version: "latest"
+
+    - name: Install dependencies with dev group
+      run: uv sync --group dev
+      working-directory: rre-dataset-generator
```
dantuzi: This code is common to all three jobs.
Have you considered the idea of having a single job with the following steps?
- Checkout
- Set up Python
- Install uv
- Install dependencies
- Run MyPy
- Run Linting (ruff)
- Run tests
The solution you proposed is clearer, and it's easier to identify which check is failing.
On the other side, the solution I described executes the first 4 steps only once

.github/workflows/test.yml
```diff
@@ -22,10 +77,10 @@ jobs:
       with:
         version: "latest"
 
-    - name: Install dependencies
-      run: uv sync
+    - name: Install dependencies with dev group
+      run: uv sync --group dev
```
dantuzi: What does `--group dev` do?

.github/workflows/test.yml
Intrinsical-AI: This triggers the execution when pushing to main or dataset-generator branches, but only when any .py file (excluding the ones insitde __pycache__), are updated / changed.

 If we remove "paths" -> On every change (for example, after changes in a .md file content, e.g: README.md) the full workflow get's executed again. Now it requires at least a change in a .py file.

.github/workflows/test.yml
Intrinsical-AI: You're right. With this approach we're executing the same commands few time. The purpose it's start and run the jobs in parallel, so we can save time and avoid waiting too much to see if the PR is passing the checks. 

I'd keep it from now, and if the space / resources consumption on the Git instance grows, maybe we need a refactor / split.

.github/workflows/test.yml
Intrinsical-AI: In the pyproject.toml we're defining sets/clusters of dependencies:

```
[project]
dependencies = [ .. ] 
```
-> This are the default, installed with pip install -e . (editable mode)


```
[dependency-groups]
dev = [ ..]
```
 -> This are the "dev" dependencies, installed with pip install -e .[dev]

So _uv sync --group dev_ install both; the defaults, and the dev ones (in editable mode)

PS: "dev" it's the common standard to use as label

.github/workflows/test.yml
nicolo-rinaldi: And also I think nobody is pushing directly on main or dataset-generator outside of PRs

.github/workflows/test.yml
```diff
@@ -1,13 +1,68 @@
...
+      uses: astral-sh/setup-uv@v3
+      with:
+        version: "latest"
+
+    - name: Install dependencies with dev group
+      run: uv sync --group dev
+      working-directory: rre-dataset-generator
+
+    - name: Run mypy
+      run: uv run mypy src/
```
nicolo-rinaldi: Just to be sure: are you checking only `src/` folder since you want to avoid tests? In that case, haven't you changed already `pyproject.toml` file to address that?

.github/workflows/test.yml
nicolo-rinaldi: It updates just the dependencies inside the dev-group in `pyproject.toml` file. I think it makes sense to update just those ones, but `pyproject.toml` must be crafted accurately to split dependencies in 2 groups (dev and non-dev). `uv sync --group dev` is faster but is a bit risky in my opinion

Intrinsical-AI: @nicolo-rinaldi Hey Nicolo, I checked again: it's not updating; it's indeed **installing** as said, and _not only the "dev" deps, but all the defaults. You can check it here:
https://docs.astral.sh/uv/concepts/projects/sync/#editable-installation

uv sync ->  **installs** the project (editable by default - I remember we discussed when the uv workspaces convo: pip install -e . / pip install -e .[dev]..)

uv sync --group `<name>` -> installs the project default dependencies, and the extra pointed by `<name>`

Note: after diving on the astral docs I realized **dev it's already consider a default by uv**. As long as we don't override  the `[tool.uv] default-groups` in the pyproject.toml we could simplify to: `uv sync`, but personally I rather have it explicit (with --group dev) for redability / missconfigurations. A possible upgrade could be adding --locked to check consitency between the build deps and the specified by uv.lock

.github/workflows/test.yml
Intrinsical-AI: I rather specify the path - there's no overhead and we're avoiding conflicts in some cases. Example: if we modify the pyproject.toml

.github/workflows/test.yml
Intrinsical-AI: Sometimes you can push directly making a hotfix for example, so I thought it could be useful. 

Also: what happen if this is in prod and someone pushes to the dataset-generator branch with a small fix? -> Branch deployed without test/checks..

 After review: there's a problem with the logic given how merges  works-> merges are compose ops containing "pushes"; so we would be duplicating the checks on the "Merge and Squash" PRs for example.
 
  I will add this logic but only for **PRs** to both main / dataset-generator. (main it's the real master, dataset-generator it's the temporary master)

Intrinsical-AI: I will add the same logic (python file changes on the /rre-dataset-generator paths), for Pull Request. I realized now it's only applied to pushes; and since we're using way more PRs, makes sense to have both)

.github/workflows/test.yml
nicolo-rinaldi: Ok, thanks for clarification

nicolo-rinaldi (approved):

.github/workflows/test.yml
dantuzi: Thank you @Intrinsical-AI

.github/workflows/test.yml
```diff
@@ -1,31 +1,107 @@
-name: Unit Tests
+name: CI Pipeline
 
 on:
   pull_request:
-    branches: [ main, dataset-generator ]
```
dantuzi, line 5: you are specifying the path but you are removing the branches where the pull request goes

.github/workflows/test.yml
```diff
@@ -1,31 +1,107 @@
...
+
+
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.sha }}
+  cancel-in-progress: true # cancel older runs when pushing several times to the same PR
+
+permissions:
+  contents: read
+  pull-requests: write # for annotations in the PR UI (e.g:mypy, ruff)
```
dantuzi: As far as I can see, mypy and ruff are not executed with the parameter `--fix` (which is correct). Why do we need the write permission for the pull request?

.github/workflows/test.yml
Intrinsical-AI: I realized if we activate in both pushes and PRs,  will be duplicating checks sometimes (for example on merges, which involves a push command). So I decided left as check for _all the PRs in all the branches_ (default when not specified).

 I would personally duplicate the logic for pushes and ensure the checks are always executed, even if redundant, but thought you might want to avoid it.  I'm not sure what  policy do we really want. Let me know, I'm a bit lost on the priorities after realizing the potentially duplicated checks👍

.github/workflows/test.yml
Intrinsical-AI: You're right! I was thinking about auto-fixes, but that's not activated neither  the purpose of this PR 👍 
Removing the  `pull-requests: write` line

Intrinsical-AI: PS: still the same, but just remembered that if some day we want to implement PRs / autofixes via API (e.g: using bots like Jules or Codex, not sure about Copilot), we would need write permisions on PRs

.github/workflows/test.yml
Intrinsical-AI: Updated

dantuzi (approved):

PR #212, opened 2025-09-08
DAGE-69: DataStore and Document bug fix

nicolo-rinaldi:
DAGE-69: [Jira ticket ](https://sease.atlassian.net/browse/DAGE-69)

Update to DataStore and Document pydantic model + main script. Now the new param in Document is keeping track of which are the document used to generate the queries and computes the score of the cartesian product pairs (doc, query) only among the flaged documents.
Other small update to llm_service.py to keep the same prompt and avoid the new field "is_used_to_generate_queries" from class Document to be sent to LLM

Executed:

ruff
mypy
unit tests
M1's main script for all 3 writers


rre-tools/commons/src/commons/data_store.py
```diff
@@ -65,9 +65,9 @@ def get_document(self, doc_id: str) -> Optional[Document]:
         """Gets a single document by its ID, or None if not found."""
         return self.docs.get(doc_id)
 
-    def get_documents(self) -> List[Document]:
+    def get_documents(self, only_cartesian_product_documents: bool = False) -> List[Document]:
         """Gets all documents."""
-        return list(self.docs.values())
+        return [doc for doc in self.docs.values() if not only_cartesian_product_documents or doc.is_used_to_generate_queries]
```
Intrinsical-AI: **[Blocking]**

Semantic missmatch -> The docstring and the behaviour. In general adding a boolean it's a simple solution, but should not change the logic / semantic and testing on a 'core' function, I think

Also, this adds a overhead: the`if not only_cartesian_product_documents` should be outside the list comprehension, and now it's checking on every pydantic object value always, even if we need all docs.

I propose instead implement this specific logic in a new  method as `get_cartesian_prod_docs` -> so we don't modify the default behaviour and we don't introduce overhead

PS: I don't see new tests / updates, but al previous checks still working. If so, I guess we would need to add new tests or upgrade / expand existing because the logic / behaviour changed and the tests _haven't noticed_. Adding the new function `get_cartesian_prod_docs` -> we wouldn't have to change the current tests for get_documents() and related, just adding new => isolating the checks "_by design_"

rre-tools/dataset-generator/config.yaml
```diff
@@ -62,7 +62,7 @@ output_format: "quepid"
 
 # For rre output format, you need other fields, e.g.:
 #id_field: "id"
-#rre_query_template: "tests/unit/resources/only_q.json"
+#rre_query_template: "dataset-generator/tests/unit/resources/only_q.json"
```
Intrinsical-AI, line 65: **[Blocking]**

⚠️ I think this is for the new repo structure, but it's not approved yet and this PR aims directly to the master (dataset-generator) branch.
 
I would wait till we approve the refactor in progress, and we manage to play a bit, test exhaustively the code, and fix the current bugs in the dataset-generator branch. I feel like in general we're (about to) experimenting a bit of mess on routing, imports, paths, git/branches missaligments..

rre-tools/dataset-generator/src/dataset_generator/main.py
```diff
@@ -65,7 +69,7 @@ def generate_and_add_queries(config: Config, data_store: DataStore, llm_service:
 def add_cartesian_product_scores(config: Config, data_store: DataStore, llm_service: LLMService) -> None:
     """Complete the (query, doc) matrix with LLM scores."""
     for query_obj in data_store.get_queries():
-        for doc_obj in data_store.get_documents():
+        for doc_obj in data_store.get_documents(only_cartesian_product_documents=True):
```
Intrinsical-AI: **[Suggestion]**
`get_cartesian_product_documents()` instead of `get_documents()`? (semantics + overhead).

rre-tools/dataset-generator/src/dataset_generator/main.py
```diff
@@ -42,15 +42,19 @@ def generate_and_add_queries(config: Config, data_store: DataStore, llm_service:
         doc_number=config.doc_number,
         doc_fields=config.doc_fields
     )
+    for doc in docs_to_generate_queries:
+        doc.is_used_to_generate_queries = True
+        data_store.add_document(doc)
```
Intrinsical-AI, line 48: **[Suggestion]**

We have the other loop in the same function:
```
 for doc in docs_to_generate_queries:
        query_response: LLMQueryResponse = llm_service.generate_queries(doc, num_queries_per_doc)
        for query_ in query_response.get_queries():
```
        
 -> To avoid 2 loops, I'd include the "new logic" there

rre-tools/dataset-generator/src/dataset_generator/main.py
```diff
@@ -42,15 +42,19 @@ def generate_and_add_queries(config: Config, data_store: DataStore, llm_service:
...
     remaining = max(0, config.num_queries_needed - len(data_store.get_queries()))
+    if remaining == 0:
+        return
     num_queries_per_doc: int = int((remaining // max(1, config.doc_number)) * 1.5)
     log.debug(f"Number of documents retrieved for generation: {len(docs_to_generate_queries)}")
     log.debug(f"Pending queries to generate: {remaining}")
     log.debug(f"Number of queries per document: {num_queries_per_doc}")
 
     for doc in docs_to_generate_queries:
-        data_store.add_document(doc)
```
Intrinsical-AI, line 52: **[Suggestion]**
We dont need the previous loop, just adding the logic here and (re) moving the log 
   ` log.debug(f"Number of documents retrieved for generation: {len(docs_to_generate_queries)}")`
   should work (?)

rre-tools/commons/src/commons/data_store.py
```diff
@@ -88,21 +88,22 @@ def get_ratings(self) -> List[Rating]:
...
+        if existing_id := self.query_text_to_query_id.get(query_text_str):
+            log.debug(f"[add_query] exists text='{query_text_str}' existing_id={existing_id}")
+            query = self.queries[existing_id]
+        else:
+            query = Query(id=query_id, text=query_text_str) if query_id else Query(text=query_text_str)
+            self.queries[query.id] = query
+            self.query_text_to_query_id[query_text_str] = query.id
+            log.debug(f"[add_query] added query_id={query.id}")
+
         return query
```
Intrinsical-AI, line 111: **[Clarification]**
This is the same as before, with a micro overhead on assigning `query = self.queries[existing_id]` (you have to save the reference if you don't return it right away); and the extra line for the `else ` xD

The query_id naming it's better as this👍

rre-tools/commons/src/commons/data_store.py
```diff
@@ -88,21 +88,22 @@ def get_ratings(self) -> List[Rating]:
     def add_document(self, doc: Document) -> None:
         """Adds a document."""
         if self.has_document(doc.id):
-            log.warning(f"[add_document] exists doc_id={doc.id}")
+            log.debug(f"[add_document] exists doc_id={doc.id}")
```
Intrinsical-AI, line 95: Nice 👍

rre-tools/commons/src/commons/data_store.py
```diff
@@ -178,7 +179,7 @@ def load(self) -> None:
         for query_as_dict in data.get("queries", []):
             try:
                 tmp_query = Query.model_validate(query_as_dict)                # Create a new tmp query with loaded dict 
-                self.add_query(query_text_str=tmp_query.text, id=tmp_query.id) # Pass (text, ID) values to keep ID consistent
+                self.add_query(query_text_str=tmp_query.text, query_id=tmp_query.id) # Pass (text, ID) values to keep ID consistent
```
Intrinsical-AI, line 186: Good upgrade on redability / semantics here

rre-tools/dataset-generator/src/dataset_generator/llm/llm_service.py
```diff
@@ -25,7 +25,7 @@ def generate_queries(self, document: Document, num_queries_generate_per_doc: int
             "Example format: [\"first query\", \"second query\"]"
         )
 
-        doc_json = document.model_dump_json()
+        doc_json = document.model_dump_json(exclude=set("is_used_to_generate_queries"))
```
Intrinsical-AI: **[Bug]**

`exclude=set("is_used_to_generate_queries")` is wrong. It applies the set to the string, not creates a set with the whole string as one element. This creates a set of characters (one character in the string- one element of the set) ⇒ it excludes nothing, but there's no error. The problem it's without testing we might it does, and we wouldnt realize till we see it -> _"Silent bug"_

Should be: `document.model_dump_json(exclude={"is_used_to_generate_queries"}) `  -  with { }

 I think it's a good practice to run the actual test and also add new specific ones (even 1 or 2 would be enough) based on the changes and the expected (new) behaviour.

rre-tools/dataset-generator/src/dataset_generator/llm/llm_service.py
```diff
@@ -87,7 +87,7 @@ def generate_score(self, document: Document, query: str, relevance_scale: str,
                 content=system_prompt
             ),
             HumanMessage(
-                content=f"Document: {document.model_dump_json()}\n"
+                content=f"Document: {document.model_dump_json(exclude=set('is_used_to_generate_queries'))}\n"
```
Intrinsical-AI: Same here. 

PS: would be top tier-level agreed on a consistant use of the " or ' as quotes overall the project (here it's different), but I guess now it's too late. Just as an idea for the next one!

Intrinsical-AI (commented):
I not consider the PR description strictly a bug, but a miss-aligment on the team expected behaviour; so I propose to reach common alignment in a meeting (with Daniele and Alessandro). I agree to change the approach as discussed today, but not sure on how exactly.

Also: we have 2 detected (confimed) bugs currently, so I'd focus on fixing them and develop more exhaustive testing, instead of adding new functionalities / changing current behaviour (even more knowing that we have a demo + presentation coming, lol)

rre-tools/dataset-generator/src/dataset_generator/llm/llm_service.py
Intrinsical-AI: POC:
image

rre-tools/dataset-generator/config.yaml
nicolo-rinaldi: new repo structure is already available, I merged it yesterday evening

rre-tools/dataset-generator/src/dataset_generator/llm/llm_service.py
nicolo-rinaldi: changed

rre-tools/dataset-generator/src/dataset_generator/llm/llm_service.py
nicolo-rinaldi: changed

rre-tools/dataset-generator/src/dataset_generator/main.py
nicolo-rinaldi: The idea is that, since now if `remaining` is 0 the function just stops before, we still need to add the document that are needed to generate the queries. So they are added beforehand with the first loop and then, in the second one, just retrieved from docs_to_generate_queries`

rre-tools/dataset-generator/src/dataset_generator/main.py
nicolo-rinaldi: This could be an idea, let me think about it

nicolo-rinaldi (commented):
I already discussed with @dantuzi and he confirmed that this is a bug and needs to be solved

rre-tools/dataset-generator/src/dataset_generator/main.py
nicolo-rinaldi: @dantuzi do you prefer to keep one function or split it in 2 different ones? Let me know

rre-tools/dataset-generator/src/dataset_generator/main.py
Intrinsical-AI: Oh I get it. We still can do something like ?
```
    remaining = max(0, config.num_queries_needed - len(data_store.get_queries()))
    if remaining == 0:
        for d in docs:
            d.is_used_to_generate_queries = True
            data_store.add_document(d)
        return
        
```
  And then also add the docs in the other loop. The difference with actual would be that we only be executing one of the loops per call to the fn

nicolo-rinaldi (commented):
Did some changes @Intrinsical-AI. Now is more compliant on what you asked

rre-tools/commons/src/commons/data_store.py
```diff
@@ -69,6 +69,10 @@ def get_documents(self) -> List[Document]:
         """Gets all documents."""
         return list(self.docs.values())
 
+    def get_cartesian_prod_docs(self) -> List[Document]:
+        """Gets only documents used to generate queries."""
+        return [doc for doc in self.docs.values() if doc.is_used_to_generate_queries]
```
Intrinsical-AI, line 74: Makes much more sense for me 👍

rre-tools/commons/tests/unit/test_data_store.py
```diff
@@ -44,6 +44,13 @@ def test_add_and_get_doc__expects__datastore_returns_the_same_document(ds, doc_a
     assert len(ds.get_documents()) == 1
     assert ds.get_document("missing-doc") is None
 
+def test_add_and_get_cartesian_product_docs__expects__datastore_returns_the_same_document(ds, doc_a, doc_b):
+    doc_a.is_used_to_generate_queries = True
+    ds.add_document(doc_a)
+    ds.add_document(doc_b)
+    assert len(ds.get_documents()) == 2
+    assert len(ds.get_cartesian_prod_docs()) == 1
```
Intrinsical-AI, line 76: **[Suggestion]**

Ok, but this is testing the happy path. I think it's useful to test "weird" paths, where it could fail, to handle unexpected scenarios. -> Trying to break the behaviour on purpose.

For example: what happens if we do 
```
doc_a.is_used_to_generate_queries = True
ds.add_document(doc_a)

doc_a.is_used_to_generate_queries = True
ds.add_document(doc_a)

```
Is consider the same doc? Gets added / duplicated, with only the changed bit? How many docs we have? And cartesian prod docs? 


This is just an example on a certain unexpect behaviour - there multiples. But in general this is how I try to design tests.

rre-tools/dataset-generator/src/dataset_generator/main.py
```diff
@@ -42,19 +42,23 @@ def generate_and_add_queries(config: Config, data_store: DataStore, llm_service:
         doc_number=config.doc_number,
         doc_fields=config.doc_fields
     )
-    for doc in docs_to_generate_queries:
-        doc.is_used_to_generate_queries = True
-        data_store.add_document(doc)
 
     remaining = max(0, config.num_queries_needed - len(data_store.get_queries()))
     if remaining == 0:
+        for doc in docs_to_generate_queries:
```
Intrinsical-AI: Like it 👍

rre-tools/dataset-generator/src/dataset_generator/main.py
```diff
@@ -69,7 +73,7 @@ def generate_and_add_queries(config: Config, data_store: DataStore, llm_service:
 def add_cartesian_product_scores(config: Config, data_store: DataStore, llm_service: LLMService) -> None:
     """Complete the (query, doc) matrix with LLM scores."""
     for query_obj in data_store.get_queries():
-        for doc_obj in data_store.get_documents(only_cartesian_product_documents=True):
+        for doc_obj in data_store.get_cartesian_prod_docs():
```
Intrinsical-AI, line 74: 👍

Intrinsical-AI (approved):
All relevant feedback has being faced succesfully 👍 

Added an optional suggestion on tests

rre-tools/dataset-generator/src/dataset_generator/main.py
Intrinsical-AI: Solved

rre-tools/commons/src/commons/data_store.py
Intrinsical-AI: Solved

rre-tools/commons/tests/unit/test_data_store.py
nicolo-rinaldi: Added couple of more tests

dantuzi (approved):

PR #216, opened 2025-09-09
DAGE-64: Refactor of rre-embeddings into rre-tools/embedding-model-evaluator

nicolo-rinaldi:
DAGE-64: [Jira ticket](https://sease.atlassian.net/browse/DAGE-64)

Main script was not working, even in the old version. Class CachedEmbeddingWrapper had some problems recognizing the tasks. It has been fixed by importing the variable TASKS_REGISTRY from mteb.overview and manually adding it after the definition of the class. In this way, the function mteb.get_task(task_name=task_name) works as expected in the whole package.

Some changes to pyproject.toml files in all packages:
- centralized common imports
- centralized test runs (to be done from rre-tools/ folder)
- mypy has one common file in rre-tools/ folder, exactly like ruff

Executed:
- ruff (that has been centralized)
- mypy (that has been centralized)
- unit test (now the ignore of DeprecationWarning happens only in embedding-model-evaluator/tests/ folder)
- main script


rre-tools/dataset-generator/pyproject.toml
```diff
@@ -1,41 +1,28 @@
 [project]
 name = "Dataset-Generator"
 version = "0.1.0"
-description = "Dataset Generator project"
+description = "Dataset Generator - A tool to generate an evaluation dataset given a collection of document"
 readme = "README.md"
+authors = []
+license = {text = "Apache-2.0"}
```
Intrinsical-AI: **[Missmatch]**

If we're adding an attribute for license I suggest referencing the real license instead of a placeholder. E.g. `"license = { file = "../../LICENSE" }"`

rre-tools/docker-services/solr-init/solr-init.sh
```diff
@@ -4,7 +4,7 @@ set -euo pipefail
 COLLECTION_ENDPOINT="http://solr:8983/solr/testcore";
 DATASET_FILE="/opt/rre-dataset-generator/data/dataset.json";
 EMBEDDINGS_FOLDER="/opt/rre-dataset-generator/embeddings";
-EMBEDDINGS_FILE="$EMBEDDINGS_FOLDER/document_embeddings.jsonl";
+EMBEDDINGS_FILE="$EMBEDDINGS_FOLDER/documents_embeddings.jsonl";
```
Intrinsical-AI: **[Typo]**

Should be `document_embeddings` as before

rre-tools/embedding-model-evaluator/config.yaml
```diff
@@ -5,23 +5,23 @@ model_id: "sentence-transformers/all-MiniLM-L6-v2"
 task_to_evaluate: "reranking"
 
 # Corpus jsonl file path with records - <id,title,text>.
-corpus_path: "resources/data/corpus.jsonl"
+corpus_path: "resources/corpus.jsonl"
```
Intrinsical-AI: **[Inconsistency]**

 Why change it? Now we're mix the embedding/ folder (also kept in resources) with raw .json files at the same level. Don't like it, I would keep the data folder.

rre-tools/mypy.ini
```diff
@@ -0,0 +1,43 @@
+[mypy]
+python_version = 3.13
```
Intrinsical-AI: **[Missmatch]**

All the references to python version have been updated to 3.10 or greater, but here

rre-tools/mypy.ini
```diff
@@ -0,0 +1,43 @@
...
+# Strict but reasonable
+disallow_untyped_defs = True
+check_untyped_defs = True
+no_implicit_optional = True
+warn_return_any = True
+show_error_codes = True
+pretty = True
+
+# Allow untyped functions in tests for speed
+[mypy-tests.*]
```
Intrinsical-AI: **[Inconsistency]**

We're excluding tests/ before, but here we're adding specific configs for them

rre-tools/pyproject.toml
```diff
@@ -3,32 +3,28 @@ name = "rre-tools"
 version = "0.1.0"
 description = "Add your description here"
 readme = "README.md"
+authors = []
+license = {text = "Apache-2.0"}
```
Intrinsical-AI: **[Plaholder]**
 If we're adding an attribute for license I suggest real content instead of a placeholder; attach a LICENSE to this folder, or link coherently. E.g. "license = { file = "../../LICENSE" }"

rre-tools/embedding-model-evaluator/pyproject.toml
```diff
@@ -0,0 +1,26 @@
+[project]
+name = "embedding_model_evaluator"
+version = "0.1.0"
+description = "Embedding Model Evaluator - A tool for embeddings exact search benchmarking"
+readme = "README.md"
+authors = []
+license = {text = "Apache-2.0"}
```
Intrinsical-AI: **[Inconsistency]**

Same here

rre-tools/pyproject.toml
```diff
@@ -3,32 +3,28 @@ name = "rre-tools"
...
-]
-
-[project.scripts]
-rre-tools = "rre_tools:main"
-
-[tool.uv.workspace]
-members = [
     "commons",
-    "dataset-generator"
+    "embedding-model-evaluator"
```
Intrinsical-AI, line 24: **[Quick Note]**

This works locally with uv workspaces -  If we ever wanted to publish the repo in pypi, we should first publish each sub-package(commons, dataset-generator, embedding-model-evaluator)

rre-tools/embedding-model-evaluator/config.yaml
```diff
@@ -5,23 +5,23 @@ model_id: "sentence-transformers/all-MiniLM-L6-v2"
...
-relevance_scale: "binary"
+relevance_scale: "graded"
 
 # (Optional) Path to write mteb output, if not given it will be written to output dir in the root folder
-output_dest: "output"
+output_dest: "resources"
 
 # (Optional) Path to write mteb document and query embeddings, if not given it will be written to output/embeddings dir
-embeddings_dest: "output/embeddings"
+embeddings_dest: "resources/embeddings"
```
Intrinsical-AI, line 26: Good 👍

rre-tools/embedding-model-evaluator/tests/unit/resources/valid_config.yaml
```diff
@@ -0,0 +1,9 @@
+model_id: "sentence-transformers/all-MiniLM-L6-v2"
+task_to_evaluate: "retrieval"
+corpus_path: "embedding-model-evaluator/tests/unit/resources/data/corpus.jsonl"
+queries_path: "embedding-model-evaluator/tests/unit/resources/data/queries.jsonl"
+candidates_path: "embedding-model-evaluator/tests/unit/resources/data/candidates.jsonl"
```
Intrinsical-AI, line 5: **[Consistency]**

Here for the tests, we're keeping the data folder [OK]. I'd keep it as before in the root of the package.

Intrinsical-AI (commented):
**[Suggestion]** 

The scope of this PR feels a bit unclear mixing configuration/setup changes, file structure updates, implementation details, naming and references changes... Is similar to what happened with the Datastore Refactor. After some review sessions with Daniele, we identified the same issue and agreed to split tasks into separate PRs next time. Doing so keeps Git history redable, cleaner, and improves teamwork and alignment overall. We can agree doing as this for the next tasks. [Source](https://www.growingdev.net/p/how-to-manage-scope-creep-in-pull). Finally: given we're already mixing scopes, it's always nice to include test or modify existing with the new behaviour.


That said, there are huge improvements in this PR around structure, configuration and centralization. Adding dedicated files for mypy and ruff configuration is solid (agree). It also fixes some previous errors/mismatches and unifies all sub-packages as independent dependencies. Overall feels like great progress 👍

rre-tools/embedding-model-evaluator/tests/unit/resources/valid_config.yaml
Intrinsical-AI: **[Extra-Suggestion]**

PS: since we're already mixing scope, what about adding some (new) lightweight tests (?)

rre-tools/docker-services/solr-init/solr-init.sh
nicolo-rinaldi: This was the name that Naz used, so I changed it accordingly

rre-tools/dataset-generator/pyproject.toml
nicolo-rinaldi: This (unfortunately) cannot be done. It throws an error during the build process of uv that happens each time I update the configuration. In your opinion, is it better to create a symlink or a copy of the file in the `rated-ranking-evaluator/` repo or just remove the licence, since the file is still in the bigger repo?

Intrinsical-AI: oh :/ 

Some tools / utilities might rely on the LICENSE (for example - publishing the lib to Pypi). I like the copy / symlink idea, could be implemented in the CI as a prev-sep / additional job . As a simpler solution just manual copy the root LICENSE (in rre-tools and also in each subdir (commons, dataset-generator..)), would work. LICENSE files does not change that much.

nicolo-rinaldi: Manual copy it is then

rre-tools/dataset-generator/pyproject.toml
```diff
@@ -1,41 +1,28 @@
 [project]
 name = "Dataset-Generator"
 version = "0.1.0"
-description = "Dataset Generator project"
+description = "Dataset Generator - A tool to generate an evaluation dataset given a collection of document"
 readme = "README.md"
+licence = {file = "LICENCE"}
```
Intrinsical-AI: 👍

rre-tools/docker-services/solr-init/solr-init.sh
Intrinsical-AI: I would still fixing the typo
image

Intrinsical-AI (approved):
All relevant feedback approached / solved

rre-tools/dataset-generator/pyproject.toml
Intrinsical-AI: @nicolo-rinaldi Hey Nicolo, just realized there's a typo: LICENCE should be LICENSE.

Also: in this case the file extension it's .md.

nicolo-rinaldi: The fact that compiled anyways is a mystery. Now is fixed

dantuzi (approved):

rre-tools/docker-services/solr-init/solr-init.sh
nseidan: @Intrinsical-AI I disagree with this renaming, 
`documents_embeddings` is the correct naming since a file contains many documents not only one document. So it should stay as before. I would rollback the change.

PR #217, opened 2025-09-11
DAGE-43: standardize query template

nicolo-rinaldi:
DAGE-43: [Jira ticket](https://sease.atlassian.net/browse/DAGE-43)

Now we give the path and validates it. If we don't provide the query_template_path param in the config.yaml file it skips the second retrieval (after the generation of the queries). rre_query_template_path is kept so we support 2 different search. If not provided, is set to the query_template_path one. If both of them are not provided and we are looking to save in RRE format, it raises and error.

Executed:
- main script
- ruff
- mypy
- tests


Intrinsical-AI (commented):
Currently, we're touching 28 files and the  [PR](https://github.com/SeaseLtd/rated-ranking-evaluator/pull/218) touches another 23, several of them in common. In general  **I like and approve the improvements**, the problem is that I’ve checked and there are potential conflicts that make the final code path hard to review, and to predict (how the conflict are gonna be solved exactly, and how the code will be like?). 

Some of the potential conflicts between both PRs include:
- dataset-generator/src/dataset_generator/config.py (breaking conflict?)
- commons/model/writer_config.py (breaks git history; needs --force or similar)
- commons/tests/unit/resources/only_q.json -> Here's but in the DAGE-66 we're removing the fixture and making the test use the direct path 'only_q.json'. I'm not sure if that's aligned.
- DAGE-66 does not expect the (new) "data" folder.

Working this way (parallel PRs affecting more total >30/40 files and including several breaking changes) we're making other working branches (including peers) outdated by default, instead trying to keep compat.

 I think here we have some opportunities to minimize the impacted files in general. For example here we're including (scope):
- Cross-compability with path bulding (nice, but not related to the PR core and impact several files)
- New structure with the /data/ folder (nice, but not related to the PR core)
- Changing names and typing (breaking) . 

**IDEA**: Why not keep the existing names query_template and query_template_rre with the new implementation, instead of adding the _path suffix? Using the previous names preserves backward compatibility, minimizes conflicts, and keeps the PR smaller. (Maybe I’m missing something, but I think there’s room to reduce impact across the files.).

--- 

I suggest either minimize the PR scope, or opening new atomic tasks with it's respective PRs, or making several PRs for the task (git history heritage/sub-branches with PRs might be the best option)

rre-tools/commons/tests/unit/writers/test_rre_writer.py
```diff
@@ -14,7 +14,7 @@ def writer_config():
         output_format='rre',
         index='testcore',
         id_field='id',
-        query_template='dataset-generator/tests/unit/resources/only_q.json',
+        query_template='commons/tests/unit/resources/only_q.json',
```
nicolo-rinaldi: This was a typo. It have been added to let tests run

nicolo-rinaldi (commented):
I removed the "data" folder inside the MtebWriter class. I'll move that fix on another ticket.

I agree on the fact that is painful to review but it's better to have meaningful names for variable, and he addition of path in the two variables you mentioned is a good way to identify the string that the user must use during the fill in of the config.yml file. 

More than this, this PR should reviewed before [DAGE-66](https://sease.atlassian.net/browse/DAGE-66) since the priority is higher in the Jira ticket. Knowing this, the plan was to rebase the main into the DAGE-66 branch to align and have the less conflict possible. I'm going to move the DAGE-66 ticket back to "in progress". I didn't think that you would start with that PR

Intrinsical-AI:
Thanks, Nicolo! Removing the `data` change from this PR makes sense.

**[Suggestion]**
On naming: **I support with clearer keys** (`query_template_path`, `rre_query_template_path`) anw, to keep backward-compat and reduce PR size/conflicts between branches while keeping the upgrade we could accept **_both_** forms in code. Examples:

```
from pydantic import ..., AliasChoices

class WriterConfig(BaseModel):
    # Accept legacy aliases: support back-compat
    query_template_path: Optional[str] = Field(
        ..,
        validation_alias=AliasChoices("query_template_path", "query_template"),
    )
    rre_query_template_path: Optional[str] = Field(
        ..,
        validation_alias=AliasChoices("rre_query_template_path", "rre_query_template"),
    )
```

```
def load_template(src: Optional[Union[str, Path]]) -> Any:
    if src is None:
        return None
    s = str(src).strip()
    if s.startswith("{") or s.startswith("["):
        return json.loads(s)          
    return Path(s).read_text("utf-8")
    
def fetch_for_evaluation(
    ...,
    query_template_path: Optional[Path] = None,
    query_template: Optional[str] = None,

):
    _ = load_template(query_template_path or query_template)
    #...
```

This should allow meaningful names + back-compat (in case we don't want to split into a different PR); preserving contracts + minimizing PR size.

PS: Not only about _review effort_! The other risks/cons are:

* **Git history** broken.
* **Unpredictable branch conflicts**.
* **Peer / parallel branches going outdated** and inheriting conflicts by default.

-> OK to deploy **DAGE-43** first and then rebase **DAGE-66** (keeping DAGE-66 focused on M3?) -
Then I’ll finalize the review and we can move to some specific/bounded PRs for polishment on namings / routes / docs / config or fixes.

 Let me know! 👍

nicolo-rinaldi:
I will do the `AliasChoices` version, but in my opinion it just delays the refactor of the files. At some point this has to be done. Half of the files of this PR are just one line change to adapt to the new version. I don't think we need backward compatibility since nothing is in production, we are still testing things out. I would agree with this approach (and I would find it very useful) if we are for example deprecating that param, so we can keep backward compatibility and maybe raise a deprecation warning, but this is not the time in my opinion. And also, since now the parameter is different, I would need to change all the files in resources folder. The param now accepts a Path (instead of a str) and even if I keep the same name, you'll find the files changed everywhere due to this fact.

Intrinsical-AI:
Thanks, Nicolo. **OK to the renaming, but not in this PR**. If changing names/types or adding aliases touches >20 files, let’s split it and keep DAGE-43 PR bounded / strict about template standardization. The code might not be production-ready yet(still thinking dataset-generator **should**), but the cost risk of Integration/CI (not only prod/CD), also matters. Finally: the main problem imo is the number of files touched (overlaps with active current PRs), not the number of lines.

**[Proposed]**

1. Land DAGE-43 with the minimal behavior (no broad renames/Path sweep).


2. Once agreed/approved, do a follow-up, focused PR for the global rename/Path changes.


3. This avoids avoidable conflicts with 5+ active branches and keeps reviews predictable.



**[Why (Fundamentals)]**

- PR-consistency: "consistency means that all the changes included on the PR should aim to solve one goal (ex. one user story) and be intrinsically related."

- PR-size: "We should always aim to keep pull requests small. Small PRs have multiple advantages: 1)They are easier to review; a clear benefit for the reviewers. 2) They are easier to deploy; this is aligned with the strategy of release fast and release often. 3) Minimizes possible conflicts and stale PRs."

- PR bounds: "Implement changes based on the well-defined description and acceptance criteria of the task at hand."

**I strongly recommend checking [Pull-Requests - Engineering Fundamentals](https://microsoft.github.io/code-with-engineering-playbook/code-reviews/pull-requests/)** and try following Git-best practices when possible.

**Note:** I think we could still changing to Path, without touching all the resource files. Union[str, Path] (config/engines) + Path(str).read_text() o json.loads() context dependant. Dummy ex: 

```
def load_template(src: Optional[Union[str, Path]]) -> Any:
    if src is None: return None
    s = str(src).strip()
    return json.loads(s) if s.startswith(("{","[")) else Path(s).read_text("utf-8")
```

**Conclusion - Proposal**
Let’s keep DAGE-43 scoped to the task, and handle the rename in a separate focused PR. This minimizes conflicts, keeps peers and tasks unblocked, aligns with recommended review and integration practices.



Thanks!

nicolo-rinaldi:
I renamed into the old ones

rre-tools/dataset-generator/tests/unit/test_solr_search_engine.py
```diff
@@ -17,7 +17,7 @@
 @pytest.fixture
 def solr_config(resource_folder):
     """Fixture that loads a valid OpenSearch config for unit tests."""
-    return Config.load(resource_folder / "solr_good_config.yaml")
+    return Config.load(resource_folder / "good_config_solr.yaml")
```
Intrinsical-AI, line 20: Nice 👍

rre-tools/dataset-generator/tests/unit/test_config.py
```diff
@@ -63,8 +63,13 @@ def test__expects__raises_file_not_found_error(resource_folder):
...
-def test_mteb_config__expects__successful_load():
-    path = "dataset-generator/tests/unit/resources/mteb_config.yaml"
-    mteb_config = Config.load(path)
+def test_mteb_config__expects__successful_load(resource_folder):
+    file_name = "mteb_config.yaml"
+    mteb_config = Config.load(resource_folder / file_name)
     assert mteb_config.output_format == "mteb"
     assert mteb_config.output_destination == Path("output")
+
+def test_missing_both_templates_with_rre__expects__raises_validation_error(resource_folder):
```
Intrinsical-AI, line 72: Like it. Nice to test this case

rre-tools/dataset-generator/src/dataset_generator/main.py
```diff
@@ -85,20 +85,25 @@ def add_cartesian_product_scores(config: Config, data_store: DataStore, llm_serv
...
+                if not data_store.has_rating_score(query_obj.id, doc_obj.id):
+                    score_resp: LLMScoreResponse = llm_service.generate_score(
+                        doc_obj, query_obj.text, config.relevance_scale, config.save_llm_explanation
+                    )
+                    data_store.create_rating_score(
+                        query_obj.id, doc_obj.id, score_resp.score,
+                        score_resp.explanation if config.save_llm_explanation else None
+                    )
+    else:
+        log.warning("Query template not found. Skipping retrieval.")
```
Intrinsical-AI, line 105: **[Suggestion]**
- Remove the Optional for config.query_template if possible.

If not possible:

1 - Using the previous query_template **would work as fallback after the warning**:
 `q={!edismax qf=title^2 description mm=1}#$query##&rows=2&fl=id,title,description,score&wt=json`
 
2- If you decide to apply it, we could simplify the code block by first instanciating a variable (e.g: `query_template_loaded`) [with ensured valid value before the if/else current block], so the execution does not get blocked. 

Not critical / pending to agree

rre-tools/dataset-generator/config.yaml
```diff
@@ -62,7 +62,7 @@ output_format: "quepid"
 
 # For rre output format, you need other fields, e.g.:
 #id_field: "id"
-#rre_query_template: "dataset-generator/tests/unit/resources/only_q.json"
+#rre_query_template: "resources/only_vector.json" # optional, if not filled, is the same as query_template_path
```
Intrinsical-AI: **[Minor]**

In the comment: same as `query_template` not  `query_template_path`. Not critical

rre-tools/commons/src/commons/model/writer_config.py
```diff
@@ -1,14 +1,14 @@
...
+from pydantic import BaseModel, Field
 
 log = logging.getLogger(__name__)
 
 class WriterConfig(BaseModel):
     output_format: Literal['quepid', 'rre', 'mteb']
     index: str = Field(..., description="Name of the index/collection of the search engine")
     id_field: Optional[str] = Field(None, description="ID field for the unique key.")
-    query_template: Optional[FilePath] = Field(None, description="Query template for rre evaluator.")
+    query_template: Optional[str] = Field(None, description="Query template for rre evaluator.")
```
Intrinsical-AI: **[Question]**

I cannot undestand this change (?) ->  FilePath extends Path (from Pathlib) and will help supporting cross-compat in the future. Being already in the master (dataset-generator), why changing the base type? Sorry, I'm a bit lost here.

rre-tools/dataset-generator/src/dataset_generator/config.py
```diff
@@ -11,7 +11,10 @@
 
 
 class Config(BaseModel):
-    query_template: str = Field("q=#$query##", description="Template string for queries with a placeholder for keywords.")
+    query_template: Optional[FilePath] = Field(
```
Intrinsical-AI, line 14: **[Question]**

Why we need it as `Optional`, and not make it required? Makes more sense to me, and would help reducing checks in higher code-layers. 

Is mandatory to have it optional, and if so why?

rre-tools/dataset-generator/src/dataset_generator/config.py
```diff
@@ -36,11 +39,20 @@ class Config(BaseModel):
     verbose: bool = False
 
     def build_writer_config(self) -> WriterConfig:
+        if self.rre_query_template is not None:
+            query_template = self.rre_query_template.name
+        else:
+            if self.query_template is not None:
+                query_template = self.query_template.name
+            else:
+                query_template = None
```
Intrinsical-AI, line 48: **[Suggestion]**

The inner if-else can be simplified to:
`query_template = self.query_template.name if self.query_template else  self.query_template`

(Not sure about the linters passing, but much more readable IMO)

rre-tools/dataset-generator/src/dataset_generator/config.py
```diff
@@ -36,11 +39,20 @@ class Config(BaseModel):
...
+
+
         return WriterConfig(
             output_format = self.output_format,
             index = self.collection_name,
             id_field = self.id_field,
-            query_template = self.rre_query_template,
+            query_template = query_template,
```
Intrinsical-AI, line 55: **[Proposal]**
To avoid using Optional in WriterConfig, we can make a single check (`query_template is not None`) before this. Not critical / pending to agree.

rre-tools/dataset-generator/src/dataset_generator/search_engine/elasticsearch_search_engine.py
```diff
@@ -71,28 +71,25 @@ def fetch_for_query_generation(self,
 
         return self._search(payload)
 
-    def fetch_for_evaluation(self, query_template: str, doc_fields: List[str], keyword: Optional[str] = None) -> List[Document]:
+    def fetch_for_evaluation(self, query_template: Path, doc_fields: List[str], keyword: Optional[str] = None) -> List[Document]:
```
Intrinsical-AI: **[Note]**

Just a note: this is not "_required_". If `fetch_for_evaluation` expects a template as str path, even suboptimal, we should keep the contract, and modify the rest of the code to pass the str value.  This is a breaking change ("refactor") that can be avoided with Union[str, Path] + casting for example. Wanted to clarify.

rre-tools/dataset-generator/src/dataset_generator/search_engine/opensearch_engine.py
```diff
@@ -61,16 +61,14 @@ def fetch_for_query_generation(self,
 
         return self._search(payload)
 
-    def fetch_for_evaluation(self, query_template: str, doc_fields: List[str], keyword: str = "*") -> List[Document]:
+    def fetch_for_evaluation(self, query_template: Path, doc_fields: List[str], keyword: str = "*") -> List[Document]:
```
Intrinsical-AI: Same here.

PS: if we had 20 engines instead of 3 (Vespa still pending to aprove), imagine how much effort we would have to do just to change a type when you call any of these functions in any part of the code.

**[Note]**
The idea to keep contracts it's that before-this-PR-invokes() still working. The concept applies to the rest of the components (not only search engines) For nexts PRs function contracts -> I would keep as defined until refactor is strictly needed (this is not the case, as we could cast before calling  or accept Union of types + after cast inside the fns)

rre-tools/dataset-generator/src/dataset_generator/search_engine/search_engine_base.py
```diff
@@ -1,14 +1,41 @@
...
 
 class BaseSearchEngine(ABC):
     def __init__(self, endpoint: HttpUrl):
         self.endpoint = HttpUrl(endpoint)
-        self.PLACEHOLDER = "#$query##"
+        self.QUERY_PLACEHOLDER = "$query"
         self.UNIQUE_KEY = 'id'
 
+
+    def _parse_query_template(self, path: Path) -> Dict[str, Any]:
```
Intrinsical-AI: **[Note]**
Related to previous reviews: back-compat simple example.
```
def _parse_query_template(self, path: Union[str|Path]) -> Dict[str, Any]:
 path = Path(path) # ensure here
```

rre-tools/dataset-generator/src/dataset_generator/search_engine/search_engine_base.py
```diff
@@ -20,7 +47,7 @@ def fetch_for_query_generation(self,
 
     @abstractmethod
     def fetch_for_evaluation(self,
-                             query_template: str,
+                             query_template: Path,
```
Intrinsical-AI: Same here

rre-tools/dataset-generator/src/dataset_generator/search_engine/solr_search_engine.py
```diff
@@ -90,22 +63,29 @@ def fetch_for_query_generation(self,
 
         return self._search(payload)
 
-    def fetch_for_evaluation(self, query_template: str, doc_fields: List[str], keyword: str="*:*") -> List[Document]:
+    def fetch_for_evaluation(self, query_template: Path, doc_fields: List[str], keyword: str="*:*") -> List[Document]:
```
Intrinsical-AI: Same here

rre-tools/dataset-generator/src/dataset_generator/config.py
Intrinsical-AI: **[Bug]**

I've realized that the main pipeline expects query_template to exist **always**. If not, it will raise an error:

Master branch current state: [commit 9d9c3ab104](https://github.com/SeaseLtd/rated-ranking-evaluator/commit/9d9c3ab104ab36e96f3c3c229fdbeea5af511fbe)

```
def expand_docset_with_search_engine_top_k(config: Config, data_store: DataStore,
                                 llm_service: LLMService, search_engine: BaseSearchEngine) -> None:
    """Retrieve docs for each query and score the (q, doc) pairs."""
    for query_obj in data_store.get_queries():
        docs_eval: List[Document] = search_engine.fetch_for_evaluation(
            keyword=query_obj.text, query_template=config.query_template, doc_fields=config.doc_fields
        )
```

There's a missmatch here. Also; the search-engines now expects a valid Path, not optional

rre-tools/commons/src/commons/model/writer_config.py
nicolo-rinaldi: We no more want validation on the path since now is just the filename, like `template.json` (as RRE expects). The validation step is one layer up, in the Config class

rre-tools/dataset-generator/config.yaml
nicolo-rinaldi: missed it, changed

rre-tools/dataset-generator/src/dataset_generator/config.py
nicolo-rinaldi: I also changed this function in the main file, so it extends to the case where query_template is None

rre-tools/dataset-generator/src/dataset_generator/config.py
nicolo-rinaldi: Query template can be None

rre-tools/dataset-generator/src/dataset_generator/main.py
nicolo-rinaldi: The behaviour that is expected is:
- if the user gives a query template file in the `config.yaml`, then the script proceed like before
- if the user doesn't give a template, then the expand_docset_with_search_engine_top_k is basically skipped, and only the scores of the dot product are generated.

rre-tools/dataset-generator/src/dataset_generator/search_engine/search_engine_base.py
nicolo-rinaldi: I added this, but the problem is that before it wasn't an actual path, but just a string with the template. In this case the path has already been validated in the during the injection of the parameter

rre-tools/dataset-generator/src/dataset_generator/config.py
nicolo-rinaldi: type checking is not passing

Intrinsical-AI (approved):

rre-tools/dataset-generator/src/dataset_generator/search_engine/elasticsearch_search_engine.py
```diff
@@ -71,28 +71,26 @@ def fetch_for_query_generation(self,
...
-        except JSONDecodeError as e:
-            raise ValueError(f"Invalid JSON query_template: {e}")
+        query_template = Path(query_template)
+        payload: Dict[str, Any] = self._parse_query_template(query_template)
+        payload = self._replace_placeholder(payload, self.QUERY_PLACEHOLDER, keyword)
 
-        query_string_obj = payload.get("query", {}).get("query_string", {})
-        if "query" in query_string_obj:
-            query_string_obj["query"] = query_string_obj["query"].replace(self.PLACEHOLDER, keyword)
+        # query_string_obj = payload.get("query", {}).get("query_string", {})
```
dantuzi, line 91: Why is this commented? If it's not needed we can directly remove it

dantuzi (approved):

rre-tools/dataset-generator/src/dataset_generator/search_engine/elasticsearch_search_engine.py
nicolo-rinaldi: I didn't see this comment, and I merged. I will remove this comments in the next PR

PR #232, opened 2025-09-25
DAGE-74: fixed bug in solr-init.sh

nicolo-rinaldi:
DAGE-74: [Jira Ticket](https://sease.atlassian.net/browse/DAGE-74)

This is a very fast review. I checked, and it behaves as described below.

The old version was indexing only the documents found in the documents_embeddings.jsonl file. The new version indexes all the documents, even if some embedding is not found. In this case, the collection will have some documents without embeddings, but all the available ones are indexed.


Intrinsical-AI (approved):
**Accepted.**  I understand this is a quick-fix for the bug. 

For the next quick-gains I propose to:
1) Refactor for encapsulating current logic into funcions with descriptive naming to improve redability, debugging, and general mantainance.
2) Add exhaustive checks and validations to avoid [breaking errors](https://sease.atlassian.net/browse/DAGE-81?atlOrigin=eyJpIjoiMGMxZmU5MzIyYzEwNDc5MDgxOWQzMmEwYzc1MTU2MjciLCJwIjoiaiJ9) outside the single-shot setup.
 
This can be done in a subsidiary task (if agreed, I will add to the backlog and implement it).


#### PS
> I think the script has grown too much in size and complexity. Future improvements might include: spliting into distinct bash files, or even better: migrating to simpler Python script via os.subprocess for example.

PR #234, opened 2025-09-25
Dage 79: (feat) Add Autosave to Datastore

Intrinsical-AI:
[TASK: DAGE-79](https://sease.atlassian.net/browse/DAGE-79)

Adds an optional auto-save feature to DataStore that periodically persists changes after N updates, enhancing data safety.

## Summary
- New config parameter: datastore_autosave_every_n_updates (default: disabled)
- Auto-save logic implementation
- Unit test expansion
-  Updated documentation

> Autosave off by default

### DIFF
--- a/rre-tools/commons/src/commons/data_store.py
+++ b/rre-tools/commons/src/commons/data_store.py
@@ -110,7 +110,7 @@ class DataStore:
         else:
             query = Query(id=query_id, text=query_text_str) if query_id else Query(text=query_text_str)
             self.queries[query.id] = query
-            self.query_text_to_query_id[query_text_str] = query.id
+            self.query_text_to_query_id[key] = query.id
             log.debug(f"[add_query] added query_id={query.id}")
             self._count_update_and_maybe_autosave()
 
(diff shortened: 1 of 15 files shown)


nicolo-rinaldi (commented):
It's better to update `test_config.py` file since you added a param to the config file
Also better to update the README file with the new param

rre-tools/commons/src/commons/data_store.py
```diff
@@ -144,6 +152,23 @@ def create_rating_score(
...
+        """Increment mutation counter and autosave if threshold reached."""
+        if self._autosave_every_n_updates is None:
+            return
+        self._updates_since_last_save += 1
+        if self._updates_since_last_save >= self._autosave_every_n_updates:
+            try:
+                self.save()
+                log.debug(f"[autosave] ok path={self.path} updates={self._updates_since_last_save}")
+            except Exception as e:
+                log.warning(f"[autosave] fail err={e}")
```
nseidan: In case of any exception, we don't save but in the finally block `self._updates_since_last_save = 0` always runs.

nseidan (approved):

PR #235, opened 2025-09-25
DAGE-84: Fix rre-tools build & Refactor READMEs

nseidan:
* Minor issues encountered while building the project. 
* jira: https://sease.atlassian.net/browse/DAGE-84


rre-tools/dataset-generator/config.yaml
```diff
@@ -54,7 +54,7 @@ num_queries_needed: 4
 relevance_scale: "graded"
 
 # Path to the LLM configuration file (e.g., LangChain setup)
-llm_configuration_file: "dataset-generator/llm_config.yaml"
+llm_configuration_file: "llm_config.yaml"
```
nicolo-rinaldi: we need dataset-generator since we start the script from the rre-tools folder

rre-tools/.gitignore
```diff
@@ -16,7 +16,7 @@ wheels/
 # Local Files and Directories
 .output_dataset
 output/
-/resources/
+resources/
```
nicolo-rinaldi: must be there, otherwise all the resource folders of tests are ignored

Intrinsical-AI: I agree, this should be `/resources` to only ignore on the root, not also in all subfolders

rre-tools/dataset-generator/config.yaml
Intrinsical-AI: I agree to run from `/rre-tools` as a standard. If so, we should include the relative route as before

Intrinsical-AI (changes requested):

rre-tools/dataset-generator/README.md
```diff
@@ -1,196 +1,98 @@
...
+Create a [config.yaml](config.yaml) file in the dataset-generator directory. This file controls the entire generation 
+process. The parameters needed are:
+> - **query_template** (Optional): Template file given by the user to evaluate the search system using relevance judgements generated by the 
+>dataset-generator module (e.g., "resources/template_solr.json")
+> - **search_engine_type**: Type of search engine to use
+>   - accepted values:
+>     - "solr"
+>     - "elasticsearch"
+>     - "opensearch"
+>     - "vespa"
```
Intrinsical-AI: Vespa not yet allowed

Intrinsical-AI (approved):
I had some minor suggestions on the Nicolos branch, but I think overall the docs are better, cleaner and cleared, so approved.

In case you want to check the minor comments/suggestions:

[REVIEW](https://github.com/SeaseLtd/rated-ranking-evaluator/pull/236#pullrequestreview-3278562234)

rre-tools/dataset-generator/README.md
nicolo-rinaldi: My bad there
@nseidan can you change it?

rre-tools/README.md
```diff
@@ -0,0 +1,100 @@
...
+Execute the main script via CLI, pointing to your DAGE configuration file:
+```bash
+uv run dataset-generator --config_file <path-to-DAGE-config-yaml>
+```
+To know more about all the possible CLI parameters, execute:
+```bash
+uv run dataset-generator --help
+```
+
+## Running Exact Search Evaluator (ESE)
```
nseidan: Rename to Embedding Model Evaluator ? @nicolo-rinaldi @Intrinsical-AI 
Also I don't prefer shortened names like DAGE or ESE ?

rre-tools/dataset-generator/README.md
nseidan: updated

nseidan: @Intrinsical-AI I looked at your comments and updated the PR, thanks.

rre-tools/README.md
nicolo-rinaldi: Yes, that was a typo on my part. I used short names to refer to the different modules, but if you don't like them, you can remove them

Intrinsical-AI: Both works for me! No strong preferences here. I would suggest only to align the folder - module naming for avoiding confusion

@nicolo-rinaldi @nseidan

nicolo-rinaldi: I'd rather stick to Embedding Model Evaluator

nseidan: updated to Embedding Model Evaluator

Intrinsical-AI: Agreed

PR #236, opened 2025-09-26
DAGE-84: Fixed READMEs and default CLI param --config

nicolo-rinaldi:


rre-tools/README.md
```diff
@@ -0,0 +1,98 @@
+# RRE tools features
+
+## Quickstart: tools installation
+
+- [uv](https://github.com/astral-sh/uv): A fast Python package installer and resolver. To installation follow instruction 
+  [here](https://docs.astral.sh/uv/getting-started/installation/)
+- Python >=3.10
```
nseidan: here we can mention,  python=3.10 version is fixed and widely used in the project,  see python-version file

rre-tools/README.md
```diff
@@ -0,0 +1,98 @@
...
+# install dependencies (editable mode - for devs)
+uv pip install -e . 
+
+# install dependencies (for users)
+uv sync
+
+# install optional dev dependencies such as mypy/ruff
+uv sync --extra dev
+```
+
```
nseidan, line 37: Add a paragraph about Search Engine that should be ready and up running or  use docker-services

rre-tools/README.md
```diff
@@ -0,0 +1,111 @@
...
+The script will then:
+1.  Fetch documents from the specified search engine.
+2.  Generate or load queries.
+3.  Score the relevance for each (document, query) pair.
+4.  (Future) Save the output to the specified destination.
+
+
+### 2. Integration Tests
+
+> INCOMING
```
nseidan: what is incoming? maybe remove, once added we can add here

rre-tools/dataset-generator/config.yaml
```diff
@@ -1,17 +1,17 @@
-# (Optional) Template string given by the user to evaluate the search system using relevance judgements generated by the
+# (Optional) Template file given by the user to evaluate the search system using relevance judgements generated by the
 # dataset-generator module
 # Default: no template is being used
 # Example (solr):
-query_template: "tests/unit/resources/template_solr.json"
+query_template: "resources/template_solr.json"
```
nseidan: isn't the `template_solr.json `in tests/unit/resources dir ?

rre-tools/docker-services/README.md
```diff
@@ -0,0 +1,106 @@
...
+
+### Running OpenSearch (Single Node)
+
+To run a local OpenSearch test environment using docker-compose:
+```bash
+cd docker-services/
+```
+
+Depending on your Docker version, you may need to use `docker compose` instead of `docker-compose`.
+If you have Docker Compose v1 installed, use:
```
nseidan: this sentence is repeated and doesn't make sense. Let's have it only once at the beginning (lines 21-22 can remain untouched and others can be removed, see also lines 71-72)

rre-tools/README.md
```diff
@@ -0,0 +1,111 @@
...
+
+## Code Quality Tools
+
+### Configuration Files
+- `ruff.toml`: Configures Ruff's linting rules and settings
+- `mypy.ini`: Configures Mypy's type checking settings
+
+### Type checker with mypy
+
+To run mypy type checks inside the dataset generator environment use
```
nseidan: no longer inside the dataset generator

rre-tools/README.md
```diff
@@ -0,0 +1,111 @@
...
+```bash
+uv run mypy .
+```
+
+### Code linter with ruff
+
+To run ruff linter inside the dataset generator environment use
```
nseidan: `dataset generator` phrase should be removed

rre-tools/dataset-generator/config.yaml
nicolo-rinaldi: I think we should not refer to some files inside the test folder in the main config. I've put all the files needed in the resource folder in the main. Do you think migh be better to push it to the repo?

nicolo-rinaldi: The downside to this is that the files are being modified often, since they are the result of the run of the different modules. We just need to align with ourselves to ignore the files inside the resources folder during reviews

nseidan: I think the config is just for us , testing purpose, why not to use tests/unit/resources?
if we add another resources folder and having files there is redundant.

nicolo-rinaldi: I was thinking of the `resources/` folder as the one where the user places all the needed files and also finds its output files (that, for example, if I run the M1, the files will be the input files for the following modules)

nseidan: I think separating input and output files is better (read and write)

nicolo-rinaldi: let's keep it the test one then

nicolo-rinaldi: @nseidan Does it make sense if I create a `templates/` folder in rre-tools? Since the templates are shared also with the 3rd module

nseidan: yes, it makes total sense

nicolo-rinaldi: done

rre-tools/README.md
```diff
@@ -0,0 +1,100 @@
+# RRE tools features
+
+## Overview
+
+### DAGE: [Dataset Generator](dataset-generator/README.md)
+
+This tool provides a flexible command-line tool to generate relevance datasets for search evaluation. It can retrieve
+documents from a search engine, generate synthetic queries, and score the relevance of document-query pairs using LLMs.
+
+### ESE: [Exact Search Evaluator](embedding-model-evaluator/README.md)
```
Intrinsical-AI, line 10: Ok I like this naming but there's a bit of missmatch with the folder name/main script invoking (embedding-model-evaluator).

 Anw, current works for me

rre-tools/README.md
```diff
@@ -0,0 +1,100 @@
...
+
+## Quickstart: tools installation
+
+- [uv](https://github.com/astral-sh/uv): A fast Python package installer and resolver. To install uv follow the 
+instructions [here](https://docs.astral.sh/uv/getting-started/installation/)
+- Python=3.10 version is fixed and widely used in the project, see [.python-version file](.python-version)
+
+First, create a virtual environment using `uv` following the file `pyproject.toml`. To do so, just execute:
+```bash
+# place yourself in the rre-tools folder
```
Intrinsical-AI, line 28: Idea: include the git clone <RRE-repo-link> before this?

rre-tools/README.md
```diff
@@ -0,0 +1,100 @@
...
+
+# install optional dev dependencies such as mypy/ruff
+uv sync --extra dev
+```
+
+## Running Dataset Generator (DAGE)
+
+Before running the command below, you need to have running search engine instance 
+(`solr`/`opensearch`/`elasticsearch`/`vespa`). This can be done even with the test collections in folder 
+[docker-services](docker-services).
```
Intrinsical-AI, line 42: If we add a README could be ./docker_services/README.md for a direct link

nseidan (approved):

rre-tools/README.md
```diff
@@ -0,0 +1,100 @@
...
+uv run embedding-model-evaluator --config <path-to-ESE-config-yaml>
+```
+
+## Running tests
+
+### 1. Unit Tests
+
+Execute `pytest` command as follows:
+```bash
+uv run pytest
```
Intrinsical-AI, line 72: Note: I might have some misconfiguration but  in Windows I had to invoke as a python module ( `uv run python -m pytest -q`) after getting the error "Failed to canonicalize the script path"-> could be mentioned as alternative

rre-tools/README.md
```diff
@@ -0,0 +1,100 @@
...
+```
+
+The script will then:
+1.  Fetch documents from the specified search engine.
+2.  Generate or load queries.
+3.  Score the relevance for each (document, query) pair.
+4.  (Future) Save the output to the specified destination.
```
Intrinsical-AI, line 79: I'm not sure about this (?)

We have MTEB writer from module 1 and DAGE-65 from M2, but not a direct wiring, I think

rre-tools/docker-services/README.md
```diff
@@ -0,0 +1,101 @@
...
+Follow the instructions to install Docker Compose on your system: https://docs.docker.com/compose/install/
+
+## Running the search engine
+
+### Running Solr (Standalone)
+
+
+To run a local Solr test environment using docker-compose:
+```bash
+cd docker-services/
```
Intrinsical-AI, line 13: maybe the whole root since the clone? 
cd rre-tools/docker-services?

rre-tools/docker-services/README.md
```diff
@@ -0,0 +1,101 @@
...
+```
+
+> This environment comes with the default Solr configuration. All text field are preprocessed as seen in the [Schemaless 
+mode](https://solr.apache.org/guide/solr/latest/indexing-guide/schemaless-mode.html) from Solr. For further info, run 
+the [Solr Admin UI](https://solr.apache.org/guide/solr/latest/getting-started/solr-admin-ui.html) and check the schema 
+of your collection.
```
Intrinsical-AI, line 19: 👍 👍 👍

rre-tools/docker-services/README.md
```diff
@@ -0,0 +1,101 @@
...
+docker compose -f docker-compose.solr.yml up --build
+```
+
+This will start 2 services:
+ - `solr`, available at http://localhost:8983/solr
+ - `solr-init`, loads documents from solr-init/data/dataset.json.
+
+**Note:** for hygiene reasons we recommend building the `Solr` container without cache first:
+```bash
+docker-compose -f docker-compose.solr.yml build --no-cache
```
Intrinsical-AI, line 37: 👍 👍 👍

rre-tools/docker-services/README.md
```diff
@@ -0,0 +1,101 @@
...
+ - `elasticsearch-init`, loads documents from elasticsearch-init/data/dataset.jsonl only if Elasticsearch doesn't have 
+any documents in the index.
+
+## Running the Quepid container
+
+```bash
+cd docker-services/
+docker compose -f docker-compose.quepid.yml up -d
+
+# then go to http://localhost/sessions/new and sign up / sign in.
```
Intrinsical-AI, line 88: I'd put only http://localhost (redirects automatically)

rre-tools/embedding-model-evaluator/README.md
```diff
@@ -1,70 +1,36 @@
...
-## ( Check the pyproject.toml line: [project.scripts] rre-embeddings = "main:main")
-rre-embeddings
-```
-
-
-### Run exact search evaluator with  yaml config file
-```bash
-rre-embeddings --config "path_to_config_yaml"
-```
+### **Input parameters** for configuration file
```
Intrinsical-AI, line 13: 👍👍👍

Intrinsical-AI (approved):

PR #228, opened 2025-09-22
Dage 21: Vespa Search Engine

Intrinsical-AI:
# [Task: DAGE-21](https://sease.atlassian.net/browse/DAGE-21?atlOrigin=eyJpIjoiODA5OTE4YmE3NzE5NDY2ODg3ZjMyNmMwNTRiNGVhMzkiLCJwIjoiaiJ9)

## Context
[Previous PR](https://github.com/SeaseLtd/rated-ranking-evaluator/pull/190)

## (features): vespa-search-engine 
- Fix proposal over the overriding fl params in the query template (found bug in Solr)
- Python class to perform search over a running instance
- Containerized (Compose + Docker + Init script) Vespa instance with document feeding.
- Bash script, intended to be run from the container; vespa init orchestration + healtchecks / validation + doc feeding
- Aim for a robust approach using quotes to avoid injection
- Added updated tests

PR #237, opened 2025-09-29
DAGE-73: MtebWriter corpus bug

nicolo-rinaldi:
DAGE-73: [Jira ticket](https://sease.atlassian.net/browse/DAGE-73)

The bug was: MtebWriter produces corpus.json with only the documents inside the datastore -> should be all

Now this problem is addressed in the `main.py` script. We are creating a new instance of the datastore with autosave DISABLED, and we are adding to that all the documents from the search engine. In this way, the corpus file is built as expected. To fetch all documents from the search engine, a new method has been implemented in the BaseSearchEngine class: `.fetch_all()`. This calls the method  `fetch_for_generation()` multiple times. This choice has been made to avoid duplication of code.

Another slight change has been made: now the default output directory for embeddings is `resources/embeddings/` instead of `output/embeddings/`.


rre-tools/dataset-generator/src/dataset_generator/main.py
```diff
@@ -149,5 +147,17 @@ def main() -> None:
...
+    # TODO:
+    #  work on a better solution, instead of adding it directly into the datastore, and maybe modify the MtebWriter
+    #  with the fetch from the search engine
+    if config.output_format == "mteb":
+        new_data_store: DataStore = DataStore()     # init again, without autosave this time (this is loading the file
+                                                    # saved in the assigned tmp folder)
+        all_doc: List[Document] = search_engine.fetch_all(doc_fields=config.doc_fields)
+        for doc in all_doc:
+            data_store.add_document(doc)
+        data_store = new_data_store
```
Intrinsical-AI: **[BUG]**

Why adding to old database, and then assign the old database to the new database?

I think with the current implementation we're always writing an empty data_store (?)


Maybe would be simpler just add to the new DS and then write the new DS in that case; data_store else.
```
if config.output_format == "mteb":
        new_data_store: DataStore = DataStore()
        all_doc: List[Document] = search_engine.fetch_all(doc_fields=config.doc_fields)
        for doc in all_doc:
            new_data_store.add_document(doc)
        writer.write(output_destination, new_data_store)
else:
        writer.write(output_destination, data_store)
```

For example?

Intrinsical-AI: If we would wanted to keep a single datastore, this might be also possible:

```
if config.output_format == "mteb":
    ids = {d.id for d in data_store.iter_documents()}
    for doc in search_engine.fetch_all(doc_fields=config.doc_fields):
        if doc.id not in ids:
            data_store.add_document(doc)
    # opcional: data_store.save() - caution to configure correcly autosave if implemented
writer.write(output_destination, data_store)
```
Anw: I think is more simpler and robust robust use 2 distinct DS instances by now and let it be till we polish the main pipeline / auxiliars [my take]

rre-tools/dataset-generator/src/dataset_generator/search_engine/search_engine_base.py
```diff
@@ -6,12 +6,32 @@
...
         self.QUERY_PLACEHOLDER = "$query"
         self.UNIQUE_KEY = 'id'
 
+    def fetch_all(self, doc_fields: List[str]) -> List[Document]:
+        """Extract all documents from search engine."""
+        # Now this is relying on fetch_for_query_generation to avoid duplicate code. Might be changed in the future
+        all_doc: List[Document] = []
+        start = 0
+        while True:
+            all_doc.extend(self.fetch_for_query_generation(documents_filter=None,
```
Intrinsical-AI: **[Robustness]**

Maybe checking

```
batch = self.fetch_for_query_generation(..)
if not batch: # avoid adding None, [], ...
        break
all_docs.extend(batch)
start += len(batch) # small improvement: len(batch) can be < DOC_NUMBER_EACH_FETCH
```

rre-tools/embedding-model-evaluator/src/embedding_model_evaluator/writers/embedding_writer.py
```diff
@@ -14,12 +14,15 @@
...
+        precision: int = 12 # it has been chosen since Solr seems to stick to 9 numbers after the comma
 ) -> None:
     with jsonlines.open(path, mode="w") as jsonl:
         for _id, vector in items:
             if isinstance(vector, np.ndarray):
-                vector = vector.tolist()
+                vector = np.round(vector, precision).tolist()
+            vector = [float(f"{x:.{precision}e}") for x in vector]
+            # format each float to a fixed precision string, then convert back to float
             jsonl.write({"id": _id, "vector": vector})
```
Intrinsical-AI, line 24: **[Suggestion]**

Cast to common type, check, and write to json avoiding float re-conversion (I belive JSON should handle as proposed; list primitive implements `_str_`)
```
arr = np.asarray(vector, dtype=np.float32)
if not np.isnan(arr) and np.isfinite(arr).all(): # Check: no n.a.ns // no infinite vals
 jsonl.write({"id": _id, "vector": arr.tolist()})
else:
 raise ...
```

rre-tools/README.md
```diff
@@ -34,8 +34,8 @@ cd rre-tools
 # install dependencies (for users)
 uv sync
 
-# install optional dev dependencies such as mypy/ruff
-uv sync --extra dev
+# install development dependencies as well (e.g., mypy and ruff)
+uv sync --group dev
```
Intrinsical-AI, line 38: **[Note]**

I think this are included in uv default profile (used with the plain uv sync). If so this is not needed but anyway I like it making it being explicit.

rre-tools/dataset-generator/tests/unit/test_elasticsearch_search_engine.py
```diff
@@ -64,6 +65,27 @@ def test_elasticsearch_search_engine_fetch_for_evaluation__expects__result_retur
                                                 doc_fields=elasticsearch_config.doc_fields)
     assert result[0] == Document(**mock_dict)
 
+def test_elasticsearch_engine_fetch_all__expects__results_returned(monkeypatch, elasticsearch_config, mock_doc, mock_dict):
```
Intrinsical-AI, line 68: Nice adding tests

rre-tools/dataset-generator/tests/unit/test_opensearch_engine.py
```diff
@@ -18,6 +19,16 @@ def opensearch_config(resource_folder):
     """Fixture that loads a valid OpenSearch config for unit tests."""
     return Config.load(resource_folder / "good_config_opensearch.yaml")
 
+@pytest.fixture
+def expected_doc():
```
Intrinsical-AI, line 23: good mock

rre-tools/embedding-model-evaluator/src/embedding_model_evaluator/config.py
```diff
@@ -18,11 +18,11 @@ class Config(BaseModel):
     candidates_path: FilePath = Field(..., description="Candidates jsonl file path")
     relevance_scale: Literal["binary", "graded"]
     output_dest: Optional[Path] = Field(
-        None, description="Path to save mteb output, by default saved in output dir."
+        None, description="Path to save mteb output, by default saved in resources dir."
     )
     embeddings_dest: Optional[Path] = Field(
         None,
-        description="Path to save mteb embeddings, by default saved in <output/embeddings> folder.",
+        description="Path to save mteb embeddings, by default saved in <resource/embeddings> folder.",
```
Intrinsical-AI: Not related // conflicts with DAGE-65 but ok let's keep

rre-tools/embedding-model-evaluator/config.yaml
```diff
@@ -19,9 +19,9 @@ candidates_path: "resources/data/candidates.jsonl"
 #   - graded: 0 (not relevant), 1 (maybe ok), 2 (that’s my result)
 relevance_scale: "graded"
 
-# (Optional) Path to write mteb output, if not given it will be written to output dir in the root folder
+# (Optional) Path to write mteb output, if not given it will be written to resources dir in the root folder
 output_dest: "resources"
 
-# (Optional) Path to write mteb document and query embeddings, if not given it will be written to output/embeddings dir
+# (Optional) Path to write mteb document and query embeddings, if not given it will be written to resources/embeddings dir
```
Intrinsical-AI, line 25: Not related // conflicts with DAGE-65

In general for this things, I'd make a general refactor PR only for comments / namings / docs; updating and cleaning.  Ok to keep as it is this time

rre-tools/dataset-generator/src/dataset_generator/search_engine/search_engine_base.py
Intrinsical-AI: Also: I realize this induces some conflicts with the [Vespa-PR](https://github.com/SeaseLtd/rated-ranking-evaluator/pull/228) (again). I've just spend some more time solving conflicts and testing. Since this can be implemented outside the base engine; are you OK moving this code to either an auxiliar file, to the main script, or somewhere else to avoid inducing conflicts? Would that be possible?

Intrinsical-AI (commented):
One potential bug / Some minor points to agree

rre-tools/dataset-generator/src/dataset_generator/main.py
nicolo-rinaldi: I will resolve with the first solution. That was a lint complaint and I changed too fast

rre-tools/embedding-model-evaluator/src/embedding_model_evaluator/writers/embedding_writer.py
nicolo-rinaldi: I already tried this. It doesn't work

Intrinsical-AI: The problem I see is that currently we're rounding 2 times. The 2nd time, when casting string -> float we pass from "rounded-but-full-prec" to "rounded-and-rerounded-again".

With expected vectors this should work, and give distinct results than current impl. (checked on Colab with json writting)
```
..
            if isinstance(vector, np.ndarray):
                vector = vector.round(decimals=precision).tolist()
            else:
                vector = [round(float(x), precision) for x in vector]
```

> If we're getting an unexpected error when casting or something like that might be corrupt data / input format missmatch or loading (None, [], np.nan, "", "'"...)

PS: the method sign currently is:
`items: Iterable[tuple[str, np.ndarray | list[float]]]`

I think it could be:
` Iterable[tuple[str, np.ndarray | Sequence[float]]] ` 

Sequence accepts wider range of types as tuples, torch.Tensor/tf.Tensor, pandas.Serie..

nicolo-rinaldi: Ok, I'll try this and change it if it works

rre-tools/dataset-generator/src/dataset_generator/search_engine/search_engine_base.py
nicolo-rinaldi: Changed

rre-tools/embedding-model-evaluator/src/embedding_model_evaluator/writers/embedding_writer.py
nicolo-rinaldi: At the end, I rolled back. The bug was no more there

nicolo-rinaldi:
Tests are now failing. Wait to review until I re-design the ones that are not correct now

rre-tools/embedding-model-evaluator/src/embedding_model_evaluator/writers/embedding_writer.py
nicolo-rinaldi: Bug actually was there, I changed the solr-init to solve only for solr

rre-tools/dataset-generator/src/dataset_generator/search_engine/search_engine_base.py
Intrinsical-AI: @nicolo-rinaldi thanks man!

Intrinsical-AI (approved):
I had this pending👍

rre-tools/dataset-generator/src/dataset_generator/search_engine/search_engine_base.py
```diff
@@ -6,12 +6,32 @@
...
         self.endpoint = HttpUrl(endpoint)
         self.QUERY_PLACEHOLDER = "$query"
         self.UNIQUE_KEY = 'id'
 
+    def fetch_all(self, doc_fields: List[str]) -> List[Document]:
+        """Extract all documents from search engine."""
+        # Now this is relying on fetch_for_query_generation to avoid duplicate code. Might be changed in the future
+        all_docs: List[Document] = []
+        start: int = 0
+        while True:
```
nseidan: while true - dangerous statement -> leads to stackoverflow

rre-tools/dataset-generator/src/dataset_generator/main.py
```diff
@@ -139,15 +138,33 @@ def main() -> None:
...
     if config.save_llm_explanation:
         if llm_explanation_path := config.llm_explanation_destination:
             data_store.export_all_records_with_explanation(llm_explanation_path)
             log.info(f"Dataset with LLM explanation is saved into: {llm_explanation_path}")
 
+    # TODO:
+    #  work on a better solution, instead of overwriting the corpus.json file, and maybe modify the MtebWriter with the
+    #  fetch from the search engine
+    if config.output_format == "mteb":
+        all_docs: List[Document] = search_engine.fetch_all(doc_fields=config.doc_fields)
```
nseidan: in case of few hundred thousands  docs in the engine, we will hit the buffer limit by querying all docs. Pass a threshold there?

nseidan (changes requested):

rre-tools/dataset-generator/src/dataset_generator/search_engine/search_engine_base.py
dantuzi: You can run the query the first time, get the total number of results (on Solr, you have the `numFound` value, in ElasticSearch/OpenSearch you have `hits.total.value`), and iterate while the start + rows < total number of documents.

Another alternative is to directly add an abstract generic `search` method in the BaseSearchEngine class that is supposed to implement the paginated search. This is extended in all SearchEngine clients, and each client implements the pagination using the specific SearchEngine API.
For example, in Solr we can only use the `start` and `rows` parameters but in ElasticSearch we can use the `search_after` parameter (https://www.elastic.co/docs/reference/elasticsearch/rest-apis/paginate-search-results).
That being said, it's probably easier to refactor the search engine classes in the future

rre-tools/dataset-generator/src/dataset_generator/main.py
```diff
@@ -139,15 +138,33 @@ def main() -> None:
...
+        corpus_path = Path(output_destination) / "corpus.jsonl"
+        with corpus_path.open("w", encoding="utf-8") as file:
+            for doc in all_docs:
+                doc_id = str(doc.id)
+                fields = doc.fields
+                title = _to_string(fields.get("title"))
+                text = _to_string(fields.get("description"))
+
+                row = {"id": doc_id, "title": title, "text": text}
+                file.write(json.dumps(row, ensure_ascii=False) + "\n")
```
dantuzi, line 170: The fields `title` and `description` are hardcoded

rre-tools/dataset-generator/src/dataset_generator/main.py
dantuzi: Instead of collecting all documents and storing them in a file, we should explore how we can implement a kind of streaming.
A possibility can be to pass a callback function to the method `fetch_all` that is invoked for every new document read from the search engine.
The implementation of the callback appends a new line to the `corpus.jsonl` file

nicolo-rinaldi: One by one is going to raise problems opening the file multiple times, I'm going to do a batch version of it

rre-tools/dataset-generator/src/dataset_generator/main.py
nicolo-rinaldi: fixed

rre-tools/dataset-generator/src/dataset_generator/main.py
nicolo-rinaldi: done

nicolo-rinaldi:
@dantuzi @nseidan I fixed what you asked. Now there are a lot more files changed, but I hope the fixes are good

rre-tools/dataset-generator/tests/unit/mocks/elasticsearch.py
```diff
@@ -2,14 +2,18 @@
...
         self._json_data = json_data
         self.status_code = status_code
+        self.total_hits = total_hits
 
     def json(self):
         return {
             "hits": {
-                "hits": self._json_data
+                "hits": self._json_data,
+                "total": {
```
nseidan, line 14: maybe use`total_hits` as you used in the other code

rre-tools/dataset-generator/src/dataset_generator/search_engine/search_engine_base.py
```diff
@@ -2,16 +2,42 @@
...
+        Args:
+            doc_fields: Fields to extract from documents
+
+        Yields:
+            List[Document]: Batch of documents
+        """
+        # Now this is relying on fetch_for_query_generation to avoid duplicate code. Might be changed in the future
+        start: int = - DOC_NUMBER_EACH_FETCH
+        total_hits: int = self._get_total_hits(self._fetch_all_payload)
+        while start + DOC_NUMBER_EACH_FETCH < total_hits:
```
nseidan: -> `start` (line#29)   is a negative number and passing it to directly `fetch_for_query_generation()`  so it will break. 
 -> not checking the batch if it is empty , in this case, it will be in an infinite loop. 
-> while statement does not handle the last page when `start + DOC_NUMBER_EACH_FETCH == total_hits`
How about starting with a simple while statement:
```
  start = 0
  page_size = DOC_NUMBER_EACH_FETCH
  total_hits: int = self._get_total_hits(...)
  while start < total_hits:
       code ...
```

Another case is about total_hits if it is changed (e.g. search engine collection/index is updated with more data, total_hits becomes stale). Maybe for now we don't consider this case

rre-tools/dataset-generator/src/dataset_generator/search_engine/solr_search_engine.py
```diff
@@ -24,10 +24,37 @@ def __init__(self, endpoint: HttpUrl):
...
+
+        # Force Solr to return a JSON formatted response
+        payload['params']['wt'] = 'json'
+
+        log.debug("Retrieving all docs to count them")
+        log.debug(f"Search url: {search_url}")
+        log.debug(f"Payload: {payload}")
+
+        try:
+            response = requests.post(search_url, headers=self.HEADERS, json=payload)
```
nseidan, line 45: why POST here for solr? Normally /select is used with GET ?

rre-tools/dataset-generator/src/dataset_generator/search_engine/vespa_search_engine.py
```diff
@@ -42,6 +42,31 @@ def __init__(self, endpoint: HttpUrl):
     # Helpers / internal utils
     # ------------------------------------------------------------------
 
+    @property
+    def _fetch_all_payload(self) -> Dict[str, Any]:
+        return {
+            'yql': f"select * from {self.schema} where true"
```
nseidan: why to use `where true` it doesn't reflect anything?

rre-tools/dataset-generator/tests/unit/mocks/elasticsearch.py
nicolo-rinaldi: Sorry, I'm not getting this comment. This should be the same of the other code

rre-tools/dataset-generator/src/dataset_generator/search_engine/vespa_search_engine.py
nicolo-rinaldi: Removed

rre-tools/dataset-generator/src/dataset_generator/search_engine/solr_search_engine.py
nicolo-rinaldi: We used POST everywhere. If we change this, we need to do it also for all other search engines

rre-tools/dataset-generator/src/dataset_generator/search_engine/search_engine_base.py
nicolo-rinaldi: Tried to fix

rre-tools/dataset-generator/src/dataset_generator/search_engine/search_engine_base.py
```diff
@@ -2,16 +2,45 @@
...
+        start: int = 0
+        total_hits: int = self._get_total_hits(self._fetch_all_payload)
+        while start < total_hits:
+            batch = self.fetch_for_query_generation(
+                documents_filter=None,
+                doc_number=DOC_NUMBER_EACH_FETCH,
+                doc_fields=doc_fields,
+                start=start
+                )
+            yield batch
```
nseidan: before yield, check if batch is empty/None, if so then `break` the loop. Otherwise, it will be in the infinite loop.

rre-tools/dataset-generator/tests/unit/mocks/elasticsearch.py
nseidan: I meant renaming it. I am seeing sometimes total_hits

nicolo-rinaldi: still not getting you. I should have written `total_hits` everywhere I'm using this (e.g., mocks.vespa.py, mocks.solr.py, mocks.elasticsearch.py, mocks.opensearch.py)

rre-tools/dataset-generator/src/dataset_generator/search_engine/search_engine_base.py
nicolo-rinaldi: done

nicolo-rinaldi: Done

nicolo-rinaldi: Done

nseidan (approved):

rre-tools/dataset-generator/src/dataset_generator/search_engine/elasticsearch_search_engine.py
```diff
@@ -37,12 +57,13 @@ def fetch_for_query_generation(self,
...
             doc_number (int): Number of documents to retrieve.
             doc_fields (List[str]): List of field names to include in the output.
+            start (int, optional): Starting index. Defaults to 0.
 
         Returns:
             List[Document]: A list of documents formatted as `Document` instances.
         """
         # Build base query
-        query: Dict[str, Any] = {"match_all": {}}
+        query: Dict[str, Any] = self._fetch_all_payload
```
dantuzi, line 66: The function _fetch_all_payload should be invoked.

```
query: Dict[str, Any] = self._fetch_all_payload()
```

rre-tools/dataset-generator/src/dataset_generator/search_engine/opensearch_engine.py
```diff
@@ -42,20 +61,20 @@ def fetch_for_query_generation(self,
...
             query = {
                 "bool": {
                     "filter": filters
                 }
             }
         else:
-            query = {
-                "match_all": {}
-            }
+            query = self._fetch_all_payload
```
dantuzi, line 72: Same as before

rre-tools/dataset-generator/src/dataset_generator/search_engine/search_engine_base.py
```diff
@@ -2,16 +2,47 @@
...
+        while start < total_hits:
+            batch = self.fetch_for_query_generation(
+                documents_filter=None,
+                doc_number=DOC_NUMBER_EACH_FETCH,
+                doc_fields=doc_fields,
+                start=start
+                )
+            if not batch:
+                break
+            yield batch
```
dantuzi: It would be simpler if we yield a single document, not a batch.
In other words, I would iterate for each document in the batch and yield each document.

rre-tools/dataset-generator/src/dataset_generator/main.py
```diff
@@ -139,15 +138,37 @@ def main() -> None:
...
+    # TODO:
+    #  work on a better solution, instead of overwriting the corpus.json file, and maybe modify the MtebWriter with the
+    #  fetch from the search engine
+    if config.output_format == "mteb":
+        # copy pasted from MtebWriter
+        corpus_path = Path(output_destination) / "corpus.jsonl"
+        corpus_path.unlink(missing_ok=True)
+        for docs_batch in search_engine.fetch_all(doc_fields=config.doc_fields):
+            with corpus_path.open("a", encoding="utf-8") as file:
+                for doc in docs_batch:
```
dantuzi: Why do you open the file for writing the single batch?
How about the following solution? (I'm considering that the fetch_all iterates on documents, not batches)

```
with corpus_path.open("a", encoding="utf-8") as file:
   for docs in search_engine.fetch_all(doc_fields=config.doc_fields):
      doc_id = str(doc.id)
      fields = doc.fields
      title = _to_string(fields.get("title"))
      text = " ".join(
      _to_string(value)
      for key, value in fields.items()
      if key != "title"
      )
      
      row = {"id": doc_id, "title": title, "text": text}
      file.write(json.dumps(row, ensure_ascii=False) + "\n")
```

rre-tools/dataset-generator/src/dataset_generator/search_engine/elasticsearch_search_engine.py
nicolo-rinaldi: since it's a property, it should not be invoked

rre-tools/dataset-generator/src/dataset_generator/search_engine/search_engine_base.py
nicolo-rinaldi: fixed

rre-tools/dataset-generator/src/dataset_generator/main.py
nicolo-rinaldi: fixed

dantuzi (approved):

PR #242, opened 2025-10-07
DAGE-66: Approximate Search Evaluator files

nicolo-rinaldi:
DAGE-66: [Jira ticket](https://sease.atlassian.net/browse/DAGE-66)

Added: 
- main script -> `pyproject.toml` changed to update the CLI command
- tests
- a lot of test files in `tests/resource/` directory

What the main script does is: Initialize the folder to be able to evaluate Elasticsearch or Solr via the mvn archetype. Then it creates the `ratings.json` file if it's not sent as input by the associated config parameter. After this, it uploads all the needed files to RRE to work on the folder declared in the `pom.xml` file generated from the archetype. At the end, the script calls the evaluation process through `mvn rre:evaluate`.


rre-tools/embedding-model-evaluator/src/embedding_model_evaluator/py.typed
nseidan, line 1: what is a `py.typed` file? And it's empty.

rre-tools/src/rre_tools/approximate_search_evaluator/config.py
```diff
@@ -0,0 +1,94 @@
...
+
+
+class Config(BaseModel):
+    query_template: FilePath = Field(
+        ...,
+        description="Path pointing to a template file for queries with a placeholder for keywords."
+    )
+    search_engine_type: Literal['solr', 'elasticsearch']
+    collection_name: str = Field(..., description="Name of the index/collection of the search engine")
+    search_engine_url: HttpUrl
```
nseidan: no description?

rre-tools/src/rre_tools/approximate_search_evaluator/config.py
```diff
@@ -0,0 +1,94 @@
...
+    search_engine_version: str = Field(default="latest", description="Search engine version.")
```
nseidan, line 22: by default, it shouldn't be latest,  it is not stable.

rre-tools/src/rre_tools/approximate_search_evaluator/config.py
```diff
@@ -0,0 +1,94 @@
...
+    )
+    ratings_path: Optional[Path] = Field(
+        None,
+        description="Path to the rre ratings file. If not given, the content of the datastore is used."
+    )
+    embeddings_folder: Optional[Path] = Field(
+        None,
+        description="Path to collect embeddings. If not given, embeddings are not collected.",
+    )
+    output_destination: Path = Field(Path("resources"), description="Path to save the output dataset.")
```
nseidan: Mention in the description will use `resources` if not specified.

rre-tools/src/rre_tools/approximate_search_evaluator/constants.py
```diff
@@ -0,0 +1,24 @@
+SOLR_SUPPORTED_VERSIONS = [
+    "4.10.4",
+    "5.5.5",
```
nseidan, line 3: not sure even if we should include 4x or 5x versions, very old..? wonder even if there are solr users of those versions?

rre-tools/src/rre_tools/approximate_search_evaluator/main.py
```diff
@@ -0,0 +1,179 @@
...
+
+log = logging.getLogger(__name__)
+
+def _parse_args() -> argparse.Namespace:
+    parser = argparse.ArgumentParser(description="Parse arguments for CLI.")
+
+    parser.add_argument(
+        "--config",
+        type=str,
+        help='Config file path to use for the application [default: "configs/approx_search_config.yaml"]',
```
nseidan: naming: maybe full name`approx_search_config.yaml` -> `approximate_search_config.yaml`

rre-tools/src/rre_tools/approximate_search_evaluator/main.py
```diff
@@ -0,0 +1,179 @@
...
+        required=False,
+        default="configs/approx_search_config.yaml",
+    )
+
+    parser.add_argument('-v', '--verbose', action='store_true',
```
nseidan, line 28: what is `store_true`? naming is very misleading

rre-tools/src/rre_tools/approximate_search_evaluator/main.py
```diff
@@ -0,0 +1,179 @@
...
+def add_vector(rating_filename: str | Path,
+               embedding_filename: str | Path,
+               datastore: DataStore) -> None:
+    """
+    Parse the writer output and add vectors
+    under the '$vector' field inside each query's placeholders.
+    """
+    log.info("Loading rating file: %s", rating_filename)
+    with open(Path(rating_filename), "r", encoding="utf-8") as f:
+        rating_data: Dict[str, Any] = json.load(f)
```
nseidan: for typing, `Dict` -> `dict` python 3.9x supports lower case type  w/o import

rre-tools/src/rre_tools/approximate_search_evaluator/main.py
```diff
@@ -0,0 +1,179 @@
...
+    return parser.parse_args()
+
+def add_vector(rating_filename: str | Path,
+               embedding_filename: str | Path,
+               datastore: DataStore) -> None:
+    """
+    Parse the writer output and add vectors
+    under the '$vector' field inside each query's placeholders.
+    """
+    log.info("Loading rating file: %s", rating_filename)
```
nseidan: log.debug mode?

rre-tools/src/rre_tools/approximate_search_evaluator/main.py
```diff
@@ -0,0 +1,179 @@
...
+    with open(Path(rating_filename), "r", encoding="utf-8") as f:
+        rating_data: Dict[str, Any] = json.load(f)
+
+    log.info("Loading embeddings from: %s", embedding_filename)
```
nseidan: log debug mode?

rre-tools/src/rre_tools/approximate_search_evaluator/main.py
```diff
@@ -0,0 +1,179 @@
...
+            query_text = placeholders.get("$query", "")
+            query_id = datastore.query_text_to_query_id.get(query_text)
+
+            if query_id and (query_id in embeddings):
+                placeholders["$vector"] = embeddings[query_id]
+                updated_queries += 1
+
+            query_dict["placeholders"] = placeholders
+
+    log.info("Updated %d queries with vectors.", updated_queries)
```
nseidan: log.debug?

rre-tools/src/rre_tools/approximate_search_evaluator/main.py
```diff
@@ -0,0 +1,179 @@
...
+    setup_logging(args.verbose)
+    config: Config = Config.load(args.config)
+
+    eval_folder = Path(f"{config.search_engine_type}-evaluator")
+    rre_resources_folder = eval_folder / "src" / "etc"
+    templates_folder = rre_resources_folder / "templates"
+    conf_sets_folder = rre_resources_folder / "configuration_sets"
+    ratings_folder = rre_resources_folder / "ratings"
+
+    if eval_folder.exists() and eval_folder.is_dir():
```
nseidan: `eval_folder.is_dir()` would return false if the folder doesn't exist. 
So eval_folder.exists() is redundant.

rre-tools/src/rre_tools/approximate_search_evaluator/main.py
```diff
@@ -0,0 +1,179 @@
...
+        shutil.rmtree(eval_folder)
+
+    setup_rre(eval_folder, config.search_engine_type, config.search_engine_version)
+
+    if rre_resources_folder.exists() and rre_resources_folder.is_dir():
```
nseidan, line 122: just `rre_resources_folder.is_dir()` is enough

rre-tools/src/rre_tools/approximate_search_evaluator/main.py
```diff
@@ -0,0 +1,179 @@
...
+                output_format='rre'
+            )
+        )
+        writer.write(ratings_folder, data_store)
+
+    if config.embeddings_folder is not None:
+        log.info("Adding vectors to ratings file...")
+        add_vector(config.query_template, config.embeddings_folder / "queries_embeddings.jsonl", data_store)
+    else:
+        log.warning("No embeddings folder was specified. If the specified templates has a '$vector' placeholder, this "
```
nseidan, line 166: it should be log.error & exception if the absense of embeddings break the RRE evaluation

rre-tools/src/rre_tools/approximate_search_evaluator/main.py
```diff
@@ -0,0 +1,179 @@
...
+                    "will break RRE evaluation.")
+
+    log.info("Running Maven RRE evaluation...")
+    run_rre_evaluate(eval_folder)
+    log.info("Evaluation finished.")
```
nseidan, line 171: in this main, almost all logs are at the info level, I think most of them should be at the debug level.

rre-tools/templates/template_elasticsearch.json
```diff
@@ -0,0 +1,10 @@
+{
+  "query": {
+    "match": {
+      "name": {
+        "query": "$query",
+        "minimum_should_match": "3<-75% 9<-85%"
```
nseidan, line 6: how did you pick those numbers ? or by default or  just guessing?

rre-tools/embedding-model-evaluator/src/embedding_model_evaluator/py.typed
nicolo-rinaldi: It's used by mypy to recognize types inside packages. If you don't put this, mypy complains because it cannot find the datatypes needed to type checking

rre-tools/src/rre_tools/approximate_search_evaluator/config.py
nicolo-rinaldi: done

rre-tools/src/rre_tools/approximate_search_evaluator/constants.py
nicolo-rinaldi: This are the ones supported by RRE. I don't see a reason to remove old versions. I know it's very unlikely someone has a solr-4.10.4, but still is supported by RRE

rre-tools/src/rre_tools/approximate_search_evaluator/main.py
nicolo-rinaldi: ok

rre-tools/src/rre_tools/approximate_search_evaluator/main.py
nicolo-rinaldi: changed

rre-tools/src/rre_tools/approximate_search_evaluator/main.py
nicolo-rinaldi: chenged

rre-tools/src/rre_tools/approximate_search_evaluator/main.py
nicolo-rinaldi: fixed

rre-tools/src/rre_tools/approximate_search_evaluator/main.py
nicolo-rinaldi: fixed

rre-tools/templates/template_elasticsearch.json
nicolo-rinaldi: from RRE archetypes

rre-tools/src/rre_tools/approximate_search_evaluator/main.py
nicolo-rinaldi: I'll write you my idea, which was the following:
- if the user decides to put a template without the placeholder `$vector` in it, the code must work either way, just not adding the vectors to the `ratings.json` file
- if the user decides to provide both embeddings of the queries and a template with the `$vector` placeholder, then everything works correctly again
- if the user provides a template with the `$vector` placeholder inside, but the embeddings are not found, the rre_tools is logging a warning and the `rre:evaluate` is going to complain afterwards

rre-tools/src/rre_tools/approximate_search_evaluator/main.py
nicolo-rinaldi: i changed a little bit

rre-tools/src/rre_tools/approximate_search_evaluator/main.py
nicolo-rinaldi: `store_true` means: when the argument is given, it's stored as true, otherwise false (that becomes the default)

rre-tools/src/rre_tools/approximate_search_evaluator/main.py
nicolo-rinaldi: I updated with a subfolder, since the `configs/` folder will be updated hopefully with all the other configs

rre-tools/src/rre_tools/approximate_search_evaluator/main.py
nicolo-rinaldi: This is a design choice, can obv be changed. I just want to discuss which is the best option

rre-tools/src/rre_tools/approximate_search_evaluator/main.py
nseidan: for now, we haven't yet decided what structure we will use. So I wouldn't place the config under /configs .

nicolo-rinaldi: where would you place it? this module doesn't have a dedicated package

rre-tools/src/rre_tools/approximate_search_evaluator/config.py
nicolo-rinaldi: "latest" here just pick the latest version **available in RRE** from solr or elasticsearch. This is done to unify elastic and solr. If we want to avoid this, I think the best way is to keep this param `search_engine_version` as mandatory, without default

rre-tools/src/rre_tools/approximate_search_evaluator/config.py
nicolo-rinaldi: changed

rre-tools/src/rre_tools/approximate_search_evaluator/main.py
nseidan: under rre-tools/approximate_search_evaluator

rre-tools/src/rre_tools/approximate_search_evaluator/main.py
```diff
@@ -0,0 +1,182 @@
...
+
+log = logging.getLogger(__name__)
+
+def _parse_args() -> argparse.Namespace:
+    parser = argparse.ArgumentParser(description="Parse arguments for CLI.")
+
+    parser.add_argument(
+        "--config",
+        type=str,
+        help='Config file path to use for the application [default: "configs/approximate_search_evaluator/default.yaml"]',
```
nseidan: better to name configs/approximate_search_evaluator_default.yaml or configs/approximate_search_evaluator.yaml

rre-tools/src/rre_tools/approximate_search_evaluator/main.py
nicolo-rinaldi: I moved it there, but I think that folder should contain code (python files), and not some configs files

rre-tools/src/rre_tools/approximate_search_evaluator/main.py
nicolo-rinaldi: then where should I put the configs files?

rre-tools/src/rre_tools/approximate_search_evaluator/main.py
nseidan: My thought is if rre:evaluate complains, then why do just warning?

rre-tools/src/rre_tools/approximate_search_evaluator/main.py
nseidan: under rre_tools/approximate_search_evaluator

rre-tools/src/rre_tools/approximate_search_evaluator/main.py
nseidan: I meant to have consistency between other packages , even this one is not a package, it will be under the named dir.

rre-tools/src/rre_tools/approximate_search_evaluator/main.py
nicolo-rinaldi: Because it's not 100% certain that RRE complains. If we don't have vectors and the don't have the `$vector` placeholder inside the template, RRE still runs but without using vectors

nseidan: so, RRE can run successfully w/o embeddings? right?

rre-tools/src/rre_tools/approximate_search_evaluator/main.py
nicolo-rinaldi: Yeah but it's not consistent if we have it mixed with python files

nicolo-rinaldi: anw, I changed it under `rre-tools/approximate_search_evaluator/`

nseidan: then create a package for this module as well?

rre-tools/src/rre_tools/approximate_search_evaluator/main.py
nicolo-rinaldi: yes, if the provided template is withuot the $vector placeholder

rre-tools/src/rre_tools/approximate_search_evaluator/main.py
nicolo-rinaldi: we need to agree on some structure then @dantuzi @nseidan

nseidan (approved):

dantuzi (approved):

PR #243, opened 2025-10-09
DAGE-95: Update  dataset

nseidan:
JIra Issue: https://sease.atlassian.net/browse/DAGE-95

Updated dataset and config, query template.


rre-tools/commons/src/commons/writers/mteb_writer.py
```diff
@@ -22,14 +22,19 @@ class MtebWriter(AbstractWriter):
...
         """
         with corpus_path.open("w", encoding="utf-8") as file:
             for doc in datastore.get_documents():
                 doc_id = str(doc.id)
                 fields = doc.fields
                 title = _to_string(fields.get("title"))
-                text = _to_string(fields.get("description"))
+                text_parts = []
+                for k, v in fields.items():
+                    if k.lower() != "id" and v is not None:
```
nicolo-rinaldi: Since mteb merges title and text before computing the embeddings, I would put something like 
```
if k.lower() != "id" and k.lower() != "title" and v is not None:
```

nicolo-rinaldi (approved):

dantuzi (approved):

PR #244, opened 2025-10-09
DAGE-96: Added solr_init.py

nseidan:
Jira issue: https://sease.atlassian.net/browse/DAGE-96
* added `solr_init.py` and removed `solr-init.sh` .
* a flag `FORCE_REINDEX`  without shutting down Solr. 

To re-run Solr with the flag:
```
docker-compose -f docker-compose.solr.yml run --rm -e FORCE_REINDEX=true solr-init
```


rre-tools/docker-services/docker-compose.solr.yml
```diff
@@ -20,8 +20,10 @@ services:
     depends_on:
       - solr
       - prepare-embeddings-dir
+    environment:
+      - FORCE_REINDEX=false
```
nicolo-rinaldi, line 24: Is ther a way to avoid changing this directly from this file? I honestly don't know. My idea was to run the docker compose with a new flag to set this force reindex, or something similar. What do you think?

nicolo-rinaldi (approved):
The code works as expected

rre-tools/docker-services/docker-compose.solr.yml
nseidan: I explicitly added this, to show we have this `FORCE_REINDEX` env var , by default it is  not enabled. So then if one wants to reindex the dataset, just pass this var via command when running  solr with docker again:
`docker-compose -f docker-compose.solr.yml run --rm -e FORCE_REINDEX=true solr-init`

dantuzi (approved):

PR #246, opened 2025-10-13
DAGE-99: Fix elasticsearch & opensearch bulk indexing with new dataset

nseidan:
Fix elasticsearch & opensearch bulk indexing with new dataset
Jira issue: https://sease.atlassian.net/browse/DAGE-99

### DIFF
--- a/rre-tools/docker-services/elasticsearch-init/elasticsearch-init.sh
+++ b/rre-tools/docker-services/elasticsearch-init/elasticsearch-init.sh
@@ -4,6 +4,7 @@ set -euo pipefail
 CLUSTER="http://elasticsearch:9200"
 INDEX="testcore"
 ENDPOINT="$CLUSTER/$INDEX"
+DATASET_PATH="/opt/rre-dataset-generator/data/dataset.jsonl"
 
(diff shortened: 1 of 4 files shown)


rre-tools/docker-services/elasticsearch-init/elasticsearch-init.sh
```diff
@@ -24,7 +25,13 @@ if ! curl -sf -XGET "$ENDPOINT"; then
     "mappings": {
       "properties": {
         "title": {"type": "text"},
-        "description": {"type": "text"}
+        "description": {"type": "text"},
```
nicolo-rinaldi: why no dynamic one? this is working just for this dataset

nicolo-rinaldi: My idea is to have both in dynamic mapping instead of just one (opensearch)

nseidan: updated

nicolo-rinaldi (approved):
For me we can merge

PR #247, opened 2025-10-14
DAGE-101: New structure updated

nicolo-rinaldi:
DAGE-101: [Jira ticket](https://sease.atlassian.net/browse/DAGE-101)

Structure:
- `rre_tools/shared/` contains just the SHARED dependencies between the 3 modules
- `rre_tools/dataset_generator/` and `rre_tools/embedding_model_evaluator/` contain the dependencies just related to that module
- config files are now in the `configs/` folder
- all READMEs related to modules, tests, and code quality tools have been moved to `docs/` folder

All the other changes are just test file and import changes


rre-tools/docs/code_quality_tools/README.md
```diff
@@ -0,0 +1,33 @@
+## Code Quality Tools
```
nseidan: i prefer to have this rre-tools/README at the root level, and remove this subdir `code_quality_tools`. 
Then in the root README file, you can mention this `/docs` dir with other READMEs.

rre-tools/docs/dataset_generator/README.md
nseidan, line 1: i'd keep naming consistent: `dataset_generator` or `dataset-generator`

rre-tools/docs/tests/README.md
```diff
@@ -0,0 +1,14 @@
+# Running tests
+
```
nseidan: why a separate README for tests. Personally, i have never seen such a structure. Is it common?

rre-tools/pyproject.toml
```diff
@@ -3,7 +3,7 @@ name = "rre-tools"
 version = "0.1.0"
 description = "Add your description here"
 readme = "README.md"
-licence = {file = "LICENSE.md"}
+license = {file = "LICENSE.md"}
```
nseidan, line 6: this file doesn't exist.

rre-tools/src/rre_tools/dataset_generator/main.py
```diff
@@ -3,27 +3,44 @@
...
 log: Logger = getLogger(__name__)
 
 
 
+def parse_args() -> argparse.Namespace:
+    parser = argparse.ArgumentParser(description='Parse arguments for CLI.')
+
+    parser.add_argument('-c', '--config', type=str,
+                        help='Config file path to use for the application [default: "configs/dataset_generator/default.yaml"]',
+                        required=False, default="configs/dataset_generator/default.yaml")
```
nseidan: Usually  just the keyword`default`  is not used . Maybe prefix_default ?

rre-tools/src/rre_tools/embedding_model_evaluator/custom_mteb_tasks/reranking_task.py
```diff
@@ -81,22 +81,33 @@ class CustomRerankingTask(AbsTaskReranking):
...
         """
         Override AbsTask.load_data. By default, AbsTask.load_data fetches datasets from the Hugging Face Hub.
         In our case, we want to use local data files (paths defined in Config), so we override this method.
         """
 
-        if config is None:
-            raise ValueError(
-                "Pass your internal Config via MTEB.run(..., config=Config)."
-            )
+        if corpus_path is None:
```
nseidan: these checks have been validated in mteb writer?

rre-tools/src/rre_tools/embedding_model_evaluator/custom_mteb_tasks/reranking_task.py
```diff
@@ -81,22 +81,33 @@ class CustomRerankingTask(AbsTaskReranking):
         },
     )
 
-    def load_data(self, config: Config, **kwargs: Any) -> None:
+    def load_data(
+            self,
+            corpus_path: Path,
+            queries_path: Path,
```
nseidan: I'd pass config as an object argument , not its values, error prone & redundant work.

nseidan (changes requested):

rre-tools/src/rre_tools/embedding_model_evaluator/custom_mteb_tasks/reranking_task.py
nicolo-rinaldi: In my opinion, the task is linked to the mteb library and not our implementation of the module. Config is strictly related to module 2 and we are passing all the parameters, where most of them are not needed to compute the actual metrics

rre-tools/src/rre_tools/dataset_generator/main.py
nicolo-rinaldi: Do you want me to change the name of the file or the description?

rre-tools/pyproject.toml
nicolo-rinaldi: What?
https://github.com/SeaseLtd/rated-ranking-evaluator/blob/DAGE-101_last-refactor/rre-tools/LICENSE.md

rre-tools/docs/dataset_generator/README.md
nicolo-rinaldi: I'm using `dataset_generator` for folders and `dataset-generator` just for the cli. Do you want me to change this?

rre-tools/docs/code_quality_tools/README.md
nicolo-rinaldi: ok

rre-tools/docs/tests/README.md
nicolo-rinaldi: Do you prefer it in the main README? For me it's ok either way

rre-tools/src/rre_tools/embedding_model_evaluator/custom_mteb_tasks/reranking_task.py
nicolo-rinaldi: Yeah, it is. You're right. Removed

rre-tools/src/rre_tools/embedding_model_evaluator/custom_mteb_tasks/reranking_task.py
nseidan: But you are passing 4 subfields of the object, why not just pass the object. It looks clean and error-free.  And we are in M2 module, not outside.

rre-tools/src/rre_tools/dataset_generator/main.py
nseidan: both.

rre-tools/docs/tests/README.md
nseidan: yes.

rre-tools/docs/dataset_generator/README.md
nseidan: then this confuses. Just pick one.

rre-tools/src/rre_tools/dataset_generator/main.py
nicolo-rinaldi: What name do you want me to use for the file? `config` might be redundant, since everything is in the `configs/` folder

rre-tools/pyproject.toml
nseidan: Github code review  modedidn't show  this file.Anyways,  I just noticed this:
`description = "Add your description here"` in the file. Can you update this?

rre-tools/src/rre_tools/embedding_model_evaluator/custom_mteb_tasks/reranking_task.py
nicolo-rinaldi: For me it's safer to pass to the `evaluation.run()` in the main function just the things that we need, avoiding overwriting some other parameters that can be defined by the mteb library, since it's captured by **kwargs

nseidan: that's why we encapsulate all fields under "config" object.  To make it atomic and pass it as a single unit. This is the standard OOP.

nicolo-rinaldi: But I think the Config class is not done for that scope. If you want to wrap the inputs to be sent to the tasks, I would prefer to do it with a dedicated class, instead of passing the whole config

rre-tools/pyproject.toml
nicolo-rinaldi: Oh yes, didn't notice that

rre-tools/docs/dataset_generator/README.md
nicolo-rinaldi: ok, changed to everything to `dataset_generator` (same also for `embedding_model_evaluator`)

nseidan: I think it's better to change it to `dataset-generator` because see other names, rated-ranking-evaluator, rre-tools etc.

rre-tools/src/rre_tools/dataset_generator/main.py
nseidan: `dataset_generator_config.yaml` -> self explanatory

rre-tools/src/rre_tools/embedding_model_evaluator/custom_mteb_tasks/reranking_task.py
nseidan: I would leave it as it was before.

rre-tools/docs/dataset_generator/README.md
nicolo-rinaldi: I need module names not to have `-`, but only `_`, to follow python guidelines

nicolo-rinaldi: About the other folders, I can change it to the `-` version if you prefer

rre-tools/src/rre_tools/dataset_generator/main.py
nicolo-rinaldi: changed

rre-tools/pyproject.toml
nseidan: I looked into the file,  in particular these classifiers: 
```
classifiers = [
    "Development Status :: 3 - Alpha",
    "Intended Audience :: Science/Research",
    "License :: OSI Approved :: Apache Software License",
    "Programming Language :: Python :: 3.10",
    "Topic :: Scientific/Engineering :: Artificial Intelligence"
]
```

I would remove 
-     "Intended Audience :: Science/Research",
-     "License :: OSI Approved :: Apache Software License"

nicolo-rinaldi: done

rre-tools/docs/dataset_generator/README.md
nseidan: If you follow python guidelines, then there is a conflict. Module names usually have an underscore, whereas package names should be with a dash.  Now you have the same name (dataset generator) with a different separator, which is confusing.  That's why i am just suggesting to simplify without any confusion.

rre-tools/src/rre_tools/embedding_model_evaluator/custom_mteb_tasks/reranking_task.py
nicolo-rinaldi: In that case, every time I need to call one of the custom tasks, we need to define a full config even if we don't use half of the parameters. I was proposing something like: 
- create a new class to wrap the things needed for the custom task
- create a property that automatically creates the class as a new field of the config (I would place this new class in the `custom_mteb_task` module, to isolate the class from the rest)
- send it to the custom task

In this way, we can test the custom task without using the config, but just this new class. Let me know what you think about this

nseidan: > every time I need to call one of the custom tasks, we need to define a full config even if we don't use half of the parameters

I assume this is for testing.If so then we can create a mock class for the Config.

nicolo-rinaldi: It is both for testing and design purposes. I was thinking about let the main and config files act as a sort of controller and be separated from the other modules

nseidan: > create a new class to wrap the things needed for the custom task

I wouldn't split config dependent on the custom task.  Config is a preliminary entry to validate all input params are valid before running the evaluation.

rre-tools/docs/dataset_generator/README.md
nicolo-rinaldi: > ok, changed to everything to `dataset_generator` (same also for `embedding_model_evaluator`)

indeed I picked the one with underscore, as mentioned here

rre-tools/src/rre_tools/embedding_model_evaluator/custom_mteb_tasks/reranking_task.py
nseidan: I wouldn't separate config from modules. It should be dependent on the module.  If main acts like a controller, then controller calls service, the service should handle all these separation things. Then we need to add a new class called service for each module, which handles configs, calling modules methods.etc.

nicolo-rinaldi: The new task can be defined both through Config (with some property, so we have validation before) and also as a separate instance. This is the only difference I would like to have

nseidan: I think main should act like an organizer , which gets all necessary things to run. It doesn't control anything on its side.

rre-tools/docs/dataset_generator/README.md
nicolo-rinaldi: Another solution can be to use `dataset-generator` everywhere but in the Python modules

rre-tools/README.md
```diff
@@ -34,8 +34,8 @@ cd rre-tools
 # install dependencies (for users)
 uv sync
 
-# install development dependencies as well (e.g., mypy and ruff)
-uv sync --group dev
+# install optional dev dependencies such as mypy/ruff
+uv sync --extra dev
```
nseidan: have you checked does this command work?

rre-tools/src/rre_tools/shared/utils.py
```diff
@@ -52,3 +52,13 @@ def clean_text(text: str) -> str:
...
     t = _WS_REGEX.sub(" ", t).strip()
     return t
+
+def join_fields_as_text(fields: dict[str, Any], exclude: list[str] | str) -> str:
+    if isinstance(exclude, str):
+        exclude = [exclude]
+
+    text_parts = []
+    for k, v in fields.items():
+        if k.lower() not in exclude and v is not None:
```
nseidan, line 62: better to pass `exclude` as a set not a  list, O(1) amortized time.

rre-tools/README.md
nicolo-rinaldi: I lost it while organizing the REAMDEs. Reverted

rre-tools/tests/embedding_model_evaluator/test_embedding_writer.py
```diff
@@ -50,20 +44,20 @@ def test_embeddings_writer_with_valid_inputs__expects__creates_jsonl_files_with_
         vectors=query_vectors
     )
 
-    config.embeddings_dest = tmp_path / "output" / "embeddings"
+    embeddings_dir = tmp_path / "output" / "embeddings"
 
     writer = EmbeddingWriter(
-        config=config,
+        corpus_path= resource_folder / "data" / "corpus.jsonl",
```
nseidan: as corpus, queries files are now placed in `resources/` so , maybe remove subdir `data` , we have `resources/corpus.jsonl` directly

nicolo-rinaldi: done

rre-tools/src/rre_tools/shared/utils.py
nicolo-rinaldi: done

nseidan (approved):

PR #250, opened 2025-10-23
DAGE-107: Add a script to download bbc news dataset and support batch indexing in docker-services

nseidan:
- Added a script to download a large bbc news dataset 
- Added support for batch indexing in docker-services solr-init.py
- Updated READMe

### DIFF
--- a/rre-tools/docker-services/scripts/extract_bbc_news_dataset.py
+++ b/rre-tools/docker-services/scripts/extract_bbc_news_dataset.py
@@ -10,15 +10,23 @@ DATASET_SIZE = 100000
 
 
 def truncate_content(txt: str) -> str:
-    if len(txt) > CONTENT_MAX_LEN:
-        truncated: str = txt[:CONTENT_MAX_LEN]
+    if len(txt) <= CONTENT_MAX_LEN:
+        return txt
 
-        next_dot = txt.find('.', CONTENT_MAX_LEN)
-        if next_dot != -1 and next_dot - CONTENT_MAX_LEN < 200:
-            truncated = txt[:next_dot + 1]
+    truncated_txt: str = txt[:CONTENT_MAX_LEN]
+    last_dot_index = truncated_txt.rfind('.')
 
-        return truncated
-    return txt
+    if last_dot_index != -1:
+        # text size <= CONTENT_MAX_LEN
+        return txt[:last_dot_index + 1]
+
+    # find the first dot after CONTENT_MAX_LEN
+    next_dot_index = txt.find('.', CONTENT_MAX_LEN)
+
+    if next_dot_index != -1:
+        return txt[:next_dot_index + 1]
+
+    return truncated_txt
 
 
 if __name__ == '__main__':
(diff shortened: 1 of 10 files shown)


rre-tools/docker-services/scripts/extract_bbc_news_dataset.py
```diff
@@ -0,0 +1,94 @@
...
+CONTENT_MAX_LEN = 1000
+DATASET_SIZE = 100000
+
+
+def truncate_content(txt: str) -> str:
+    if len(txt) > CONTENT_MAX_LEN:
+        truncated: str = txt[:CONTENT_MAX_LEN]
+
+        next_dot = txt.find('.', CONTENT_MAX_LEN)
+        if next_dot != -1 and next_dot - CONTENT_MAX_LEN < 200:
```
nicolo-rinaldi: In this case, if we find a dot after the max len can become 1200, right?
Maybe we can check the last dot backward, so we keep the content len under `CONTENT_MAX_LEN`

rre-tools/src/rre_tools/embedding_model_evaluator/main.py
```diff
@@ -101,14 +102,17 @@ def main() -> None:
...
     evaluation = mteb.MTEB(tasks=[task])
     evaluation.run(
         model=model_with_cache,
         output_folder=config.output_dest,
         overwrite_results=True,
         config=config,
     )
+    end = time.time()
     log.info("Finished MTEB evaluation.")
+    print(f"Time took for MTEB evaluation: {end - start:.2f} seconds")
```
nicolo-rinaldi: I find this very useful to be added to logs

rre-tools/docker-services/solr-init/solr_init.py
```diff
@@ -32,7 +33,7 @@
 log = logging.getLogger("solr_init")
 
 
-def wait_for_solr_core(endpoint: str, timeout: int = 30, interval: float = 1.0) -> None:
+def wait_for_solr_core(endpoint: str, timeout, interval: float = 1.0) -> None:
```
nicolo-rinaldi: type for `timeout` variable is missing

nicolo-rinaldi (approved):
Sorry for the typos in the README and thanks for spotting them. I left minor comments

rre-tools/docker-services/scripts/extract_bbc_news_dataset.py
nseidan: right, its max is 1200, because the reason is i noticed there is a long sentence  which sometimes reaches at least 1100 and in case if we remove the sentence which reaches the last sentence after reaching 1000 chars and the content becomes around 800. That's why I set upper bound to 1200. But on average I saw about 1100 chars per content. Anyways, it is changeable.

rre-tools/src/rre_tools/embedding_model_evaluator/main.py
nseidan: my bad, it should  be log.info not print.

rre-tools/docker-services/solr-init/solr_init.py
nseidan: updated

rre-tools/src/rre_tools/embedding_model_evaluator/main.py
nseidan: updated

rre-tools/docker-services/scripts/extract_bbc_news_dataset.py
nicolo-rinaldi: For me is fine like it is right now, indeed I approved. So feel free to keep it like this. The only thing is that when I read the code, `CONTENT_MAX_LEN` is set to 1000 but the "actual one" is 1200 (i.e., I often end up with more than 1k chars), since we are addressing the problem of finding the first dot after 1000. But this is just a little thing. I don't want to block anything

rre-tools/docker-services/scripts/extract_bbc_news_dataset.py
```diff
@@ -0,0 +1,94 @@
...
+    return txt
+
+
+if __name__ == '__main__':
+
+    parser = argparse.ArgumentParser(description="Extract BBC News Dataset")
+    parser.add_argument('--filename', type=str, default='dataset.json', help='Output filename')
+    args = parser.parse_args()
+
+    months = ['2025-06', '2025-05', '2025-04', '2025-03', '2025-02', '2025-01',
```
dantuzi: How about something like
```
months = [
   f"{date.year}-{date.month}"
   for date in range (<something from X to Y increasing every month>)
]

nseidan: updated

rre-tools/docker-services/scripts/extract_bbc_news_dataset.py
nseidan: updated the code for readability

dantuzi (approved):

PR #252, opened 2025-10-30
DAGE-102: External Solr plugin implementation for version 9.8.1

nicolo-rinaldi:
DAGE-102: [Jira ticket](https://sease.atlassian.net/browse/DAGE-102)

Update External Solr plugin to version 9.8.1 -> no more dependencies on restlet -> no more compilation problems since restlet version needed by solr 8.3.0 was deprecated.

Fixed bug in logging the number of queries that needs to be executed: before, with more than one version, it happened something like "...completed -10 / 10 ..." queries, now is fixed.

After SolrJ 9.3.x, the class HttpSolrClient has been depracated, so the need to change into Http2SolrClient. The change has been made also for CloudHttp2SolrClient one, to be aligned with the rest of the code. 

A lot of bug raised, due to aynchronism and parallelism of task. Things done to solve:
1. added to counters `submittedQueries` and `completedQueries`.
2. added synchronized to the function `initialiseVersions`, to solve Concurrency exceptions.
3. `isRunning` function changed accordingly to use new counters from point 1, to check the actual retrieval of all the data needed to proceed with the computation of the resulting json file (serialization).

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -32,7 +32,7 @@
         <maven.compiler.target>1.8</maven.compiler.target>
         <jackson.version>2.9.4</jackson.version>
         <elasticsearch.version>7.5.0</elasticsearch.version>
-        <solr.version>8.3.0</solr.version>
+        <solr.version>9.8.1</solr.version>
         <github.global.server>github</github.global.server>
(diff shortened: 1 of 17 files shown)


agazzarini:
Looks good to me!

agazzarini (approved):

PR #253, opened 2025-10-31
DAGE-115: RRE Vector Search Doctor refactor

nicolo-rinaldi:
DAGE-115: [Jira ticket](https://sease.atlassian.net/browse/DAGE-115)


nseidan:
@nicolo-rinaldi how about mentioning the vector search doctor in the root README ? saying that we have the vector search doctor which has two model evaluators inside it. So that it gives an organized flow.

nseidan (approved):

PR #254, opened 2025-10-31
DAGE-110: nDCG formula update in RRE code

nicolo-rinaldi:
DAGE-110: [Jira ticket](https://sease.atlassian.net/browse/DAGE-110)

This PR is on top of [DAGE-102](https://github.com/SeaseLtd/rated-ranking-evaluator/pull/252). Better to merge the other one first. 

Related to this PR: 
- changes to NDCGAtK: implemented same metric to MTEB, with the first formula [here](https://en.wikipedia.org/wiki/Discounted_cumulative_gain).
- changes to tests related to NDCGAtK and NDCGAtTen, Fixed where needed the expected value of the metric

### DIFF
--- a/rre-core/src/main/java/io/sease/rre/core/domain/metrics/impl/NDCGAtK.java
+++ b/rre-core/src/main/java/io/sease/rre/core/domain/metrics/impl/NDCGAtK.java
@@ -108,7 +108,7 @@ public class NDCGAtK extends Metric {
                                 dcg = numerator;
                             } else {
                                 double den = Math.log(rank + 1) / Math.log(2);
-                                dcg = dcg.add(numerator.divide(new BigDecimal(den), 2, RoundingMode.FLOOR));
+                                dcg = dcg.add(numerator.divide(new BigDecimal(den), 5, RoundingMode.FLOOR));
                             }
                         });
             }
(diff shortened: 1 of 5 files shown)


rre-core/src/main/java/io/sease/rre/core/domain/metrics/impl/NDCGAtK.java
```diff
@@ -102,12 +102,13 @@ public void collect(final Map<String, Object> hit, final int rank, final String
                 judgment(id(hit))
                         .ifPresent(judgment -> {
                             final BigDecimal value = gainOrRatingNode(judgment).map(JsonNode::decimalValue).orElse(fairgrade);
-                            BigDecimal numerator = BigDecimal.valueOf(Math.pow(TWO.doubleValue(), value.doubleValue())).subtract(BigDecimal.ONE);
+                            //BigDecimal numerator = BigDecimal.valueOf(Math.pow(TWO.doubleValue(), value.doubleValue())).subtract(BigDecimal.ONE);
```
nseidan: why comment out? instead of removing?

nseidan (approved):

rre-core/src/main/java/io/sease/rre/core/domain/metrics/impl/NDCGAtK.java
nicolo-rinaldi: To keep the old implementation. I don't like it honestly. Do you think is better to remove it?

nseidan: I would rather remove it, github history is enough if ever wanted to rollback

nicolo-rinaldi: removed

PR #258, opened 2025-11-05
DAGE-104: Add info/debug level logging

nseidan:
* added some logs (info/debug level)

### DIFF
--- a/rre-tools/src/rre_tools/dataset_generator/main.py
+++ b/rre-tools/src/rre_tools/dataset_generator/main.py
@@ -26,7 +26,6 @@ from rre_tools.dataset_generator.config import Config
 log: Logger = getLogger(__name__)
 
 
-
 def parse_args() -> argparse.Namespace:
     parser = argparse.ArgumentParser(description='Parse arguments for CLI.')
 
@@ -40,7 +39,6 @@ def parse_args() -> argparse.Namespace:
     return parser.parse_args()
 
 
-
 def add_user_queries(config: Config, data_store: DataStore) -> None:
     """Loads queries from file (if exists) and adds them as Query objects."""
     if config.queries is not None:
@@ -49,6 +47,7 @@ def add_user_queries(config: Config, data_store: DataStore) -> None:
                 clean_line = line.strip()
                 if clean_line:
                     data_store.add_query(clean_line)
+            log.info(f"Added user-defined queries from file={config.queries}")
 
 
 def generate_and_add_queries(config: Config, data_store: DataStore, llm_service: LLMService,
@@ -74,7 +73,8 @@ def generate_and_add_queries(config: Config, data_store: DataStore, llm_service:
     log.debug(f"Number of queries per document: {num_queries_per_doc}")
 
     for doc in docs_to_generate_queries:
-        query_response: LLMQueryResponse = llm_service.generate_queries(doc, num_queries_per_doc, config.max_query_terms)
+        query_response: LLMQueryResponse = llm_service.generate_queries(doc, num_queries_per_doc,
+                                                                        config.max_query_terms)
         for query_ in query_response.get_queries():
             if len(data_store.get_queries()) >= config.num_queries_needed:
                 return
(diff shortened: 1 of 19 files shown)


rre-tools/src/rre_tools/dataset_generator/main.py
```diff
@@ -149,6 +149,7 @@ def main() -> None:
 
     # score initial docset
     if config.enable_cartesian_product:
+        log.debug("Cartesian product is enabled, so adding cartesian product scores")
```
nicolo-rinaldi: I would put this inside its function

rre-tools/src/rre_tools/vector_search_doctor/embedding_model_evaluator/main.py
```diff
@@ -83,14 +83,13 @@ def _get_mteb_leaderboard_avg_main_score(model_name: str, task_type: str) -> flo
...
         for task_result in model_results.task_results:
-            for split_name, split_subsets in task_result.scores.items():
-                for task_subset in split_subsets:
-                    if "main_score" in task_subset:
-                        val = task_subset["main_score"]
-                        if isinstance(val, (int, float)):
-                            scores.append(float(val))
+            val = task_result.get_score()
+            if isinstance(val, (int, float)):
+                scores.append(float(val))
```
nicolo-rinaldi, line 88: Is this a change in the logic?

rre-tools/src/rre_tools/vector_search_doctor/embedding_model_evaluator/main.py
```diff
@@ -120,8 +120,13 @@ def main() -> None:
         log.error("Custom task name is not defined: %s", task_name)
         raise ValueError("Custom task name is not defined.")
 
+    if config.output_dest is None:
```
nicolo-rinaldi, line 123: I think this is never `None`. The `Config` class for Embedding Model Evaluator is dealing with that, so it defaults to `resources/` anyway. Maybe we can log the fact that the `ourput_dest` is set to resource in the `config.py` file

nicolo-rinaldi (commented):
Does it make sense to:
- change the message logged in the `dataset_generator/config.py`?
- log the payloads/url used in the requests?

rre-tools/src/rre_tools/vector_search_doctor/embedding_model_evaluator/main.py
nseidan: it is never set to None, it is to comply with ruff/type check

rre-tools/src/rre_tools/vector_search_doctor/embedding_model_evaluator/main.py
nseidan: here using the `get_score()` method directly from task result instead of going each subset task result.

rre-tools/src/rre_tools/vector_search_doctor/embedding_model_evaluator/main.py
nicolo-rinaldi: Sad, ok

nseidan:
> change the message logged in the dataset_generator/config.py?

what messages you want to change? 

> log the payloads/url used in the requests?

I think we should not log the entire payload, because it might be lengthy, if we log at the debug level, we shoudl truncate it. 
But we should not log at the info level, because of security reason. Currently we dont support redacted ones. Personally i have never seen payload to be logged, unless to  log particular fields.

nicolo-rinaldi:
I would change [this](https://github.com/SeaseLtd/rated-ranking-evaluator/blob/163e0eaa63aa2c752d2e9f2c6c7331f3650b5921/rre-tools/src/rre_tools/dataset_generator/config.py#L155) into something that says that the config file is the one related to the dataset generator, as we did for the other 2 modules.

For logging the payload, I agree to log it in debug mode (even before, but I didn't specify it). I just think it is very useful when debugging the code to have the payload/url printed, so I can see what the code is actually doing.

rre-tools/src/rre_tools/dataset_generator/main.py
nseidan: updated

nseidan:
partial payload is (first 500chars) logged at the debug level

nicolo-rinaldi (approved):

PR #259, opened 2025-11-13
DAGE-122: added escape function to SolrSearchEngine and applied when retrieving doc using the template

nicolo-rinaldi:
DAGE-122: [Jira ticket](https://sease.atlassian.net/browse/DAGE-122)

Reference for the function: [SolrQueryParserBase.html#escape](https://github.com/apache/solr/blob/98e30eee8f0107381d9b1210fdf33aaeeb8008f5/solr/core/src/java/org/apache/solr/parser/SolrQueryParserBase.java#L992)

Implemented the function to add an escape to special characters and avoid having incongruence due to wildcards. Added also tests to keep consistency.

### DIFF
--- a/rre-tools/src/rre_tools/shared/search_engines/solr_search_engine.py
+++ b/rre-tools/src/rre_tools/shared/search_engines/solr_search_engine.py
@@ -19,8 +19,6 @@ class SolrSearchEngine(BaseSearchEngine):
     """
     Solr implementation to search into a given collection
     """
-    SPECIAL_CHARS: set[str] = {'\\', '+', '-', '!', '(', ')', ':', '^', '[', ']', '"',
-                     '{', '}', '~', '*', '?', '|', '&', '/'}
 
     def __init__(self, endpoint: HttpUrl):
         super().__init__(endpoint)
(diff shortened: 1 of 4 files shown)


rre-tools/src/rre_tools/shared/search_engines/solr_search_engine.py
```diff
@@ -27,11 +29,21 @@ def __init__(self, endpoint: HttpUrl):
...
 
+
     @property
     def _fetch_all_payload(self) -> Dict[str, Any]:
         return {
             'q': '*:*',
         }
+    @staticmethod
+    def escape(string: str) -> str:
+        """Escape special characters used in query syntax."""
```
nseidan: this escape method should be part of `util` class, applicable to other engines, because parsing is done at the lucene level, not solr.

nicolo-rinaldi: I found this as a method inside the SolrQueryParserBase class. That's why I added it to our Solr class

nseidan: yeahh, but SolrQueryParserBase is a helper class part of parser not the main one where search is made.

nicolo-rinaldi: I agree it's a helper, but I think it is related to Solr only in our case. I was planning in the future to see if there is the same problem for the other search engines and create an escape function for each one. That's why I placed it inside the Solr class. I don't know if other search engines have different rules to escape.

nseidan: lucene is the common ground for other  2 (elastic + opensearch) engines, and parser is coming from lucene.

nicolo-rinaldi: I can put it in the SearchEngineBase class, and then we override for Vespa. What do you think?

nseidan: sounds good!

nicolo-rinaldi: Done

nseidan (approved):

